PR #4, opened 2024-08-04
fixed positioning on first launch for macbook14(or all idk)

wwuselessww:
i fixed weird mispositioning on first hover. it works on macbook pro 14 on macos 14.2.1


iamharshdev:
Thanks for this, need to test before i merge and suggest changes if needed.

PR #19, opened 2024-08-07
Refactor, settings, colored spectrograms, UI changes

richardkunkli:
# Fixes
- Refactor BoringViewModel observed object to initialize in the appdelegate. It is important because the new settings changes variables of the env object and it has to be in the same instance
- Changed icon in the HeaderExtras, because it was confusing
# New features
- Implemented a new extension that extract the average color from the album art, and it is used in the audio spectrogram, which has been optimized and resized
- New settings window with some options, giving users ability to customize their experience
# Preview
Screenshot 2024-08-07 at 20 41 51

### DIFF
--- a/boringNotch/components/BoringNotch.swift
+++ b/boringNotch/components/BoringNotch.swift
@@ -29,20 +29,22 @@ struct BoringNotch: View {
                     BoringHeader(vm: vm, percentage: batteryModel.batteryPercentage, isCharging: batteryModel.isPluggedIn).padding(.leading, 6).padding(.trailing, 10).animation(.spring(response: 0.7, dampingFraction: 0.8, blendDuration: 0.8), value: vm.notchState)
                 }
                 
-                HStack(spacing: 10) {
+                HStack(spacing: 15) {
                     if vm.notchState == .closed && batteryModel.showChargingInfo {
                         Text("Charging")
                     }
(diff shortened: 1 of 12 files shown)


iamharshdev:
Thanks for @richardkunkli this but please make sure to add some preview here 🔥

richardkunkli:
Noted. Added a preview.

iamharshdev:
🔥 Loved the changes!

boringNotch/components/BoringNotch.swift
```diff
@@ -99,25 +101,21 @@ struct BoringNotch: View {
...
-                    
                     if vm.notchState == .closed && batteryModel.showChargingInfo {
                         BoringBatteryView(batteryPercentage: batteryModel.batteryPercentage, isPluggedIn: batteryModel.isPluggedIn)
                     }
                     
-                    if musicManager.isPlaying && !batteryModel.showChargingInfo && vm.currentView != .menu {
+                    if !batteryModel.showChargingInfo && vm.currentView != .menu {
                         
-                        MusicVisualizer()
+                        MusicVisualizer(avgColor: musicManager.albumArt.averageColor(), isPlaying: musicManager.isPlaying)
```
iamharshdev, line 110: You should never do extensive tasks on application thread. @richardkunkli 
I'll fix this no worries.

PR #36, opened 2024-08-12
Feature/fix multiple screen

Sn0wBiT:
https://github.com/user-attachments/assets/5b2b61a3-7cfc-4b53-a878-b118c532ae62


iamharshdev:
Looks good but needs couple more changes.

MusicVisualizer component in boringnotch file needs padding to be removed. line number 200 maybe.

line 245 : let closedWidth: CGFloat = vm.sizes.size.closed.width! - 10 change from - 20 to - 10

@Sn0wBiT make these changes else looks good to me.

PR #55, opened 2024-08-14
Fix Play/Pause Glitch in MusicManager by Implementing Debouncing

cal1b4n:
## Problem:
Rapidly pressing the play/pause button in the MusicManager could cause the music player to become glitchy or unresponsive. This issue was due to race conditions and inconsistent state updates caused by rapid successive calls to the togglePlayPause function.
## Solution:
Implemented a debounce mechanism in the togglePlayPause function to prevent it from being triggered too frequently in a short period.
Ensured thread safety by making sure that all state changes related to music playback are executed on the main thread, reducing the likelihood of race conditions.
Updated the musicIsPaused and fetchNowPlayingInfo methods to ensure consistent playback state management.
## Changes:
Added a DispatchWorkItem to debounce the togglePlayPause function.
Wrapped state updates in DispatchQueue.main.async to ensure they run on the main thread.
Improved the logic to check if the music player is in a paused state before updating the playback state.
## Testing:
Manually tested the play/pause button with rapid presses to verify that the glitch is resolved.
Verified that other functionalities of the MusicManager (such as fetching now playing info, updating album art, and handling playback state) continue to work as expected.
## Checklist:
 - Implemented debounce for play/pause toggle.
 - Ensured thread safety in state updates.
 - Manually tested the changes to confirm the issue is resolved.
## Related Issues:
I also see that Album Artworks and Artists + Song Titles are being changed correctly after implementing this.

Please do check on your side as well.

With kind regards,
Saba.

### DIFF
--- a/boringNotch/managers/MusicManager.swift
+++ b/boringNotch/managers/MusicManager.swift
@@ -1,3 +1,11 @@
+//
+//  PlaybackManager.swift
+//  boringNotch
+//
+//  Created by Harsh Vardhan  Goswami  on  04/08/24.
+//
+
+
 import SwiftUI
 import Combine
 import AppKit
--- a/boringNotch/managers/MusicManager.swift
+++ b/boringNotch/managers/MusicManager.swift
@@ -1,15 +1,8 @@
-    //
-    //  MusicManager.swift
-    //  boringNotch
-    //
-    //  Created by Harsh Vardhan  Goswami  on 03/08/24.
-    //
-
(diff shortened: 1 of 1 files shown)


iamharshdev:
The only problem is when consicutive songs have same album art, just replace the condition in music player where we check for if the album art is same as previous one we can keep that condition at last just to keep the title change and rest of othet things change. @cal1b4n

PR #83, opened 2024-08-29
Update the notch height on display event

TinyTanic:
Hi,
I plug and unplug an external monitor a lot. My primary monitor is the external one so the notch so if i start the application when the external monitor is plugged in when i unplug it it's too small on the mac monitor and viceversa.

I'm not a swift developer and i never programmed an application for mac so i don't know the best practices and the architecture for the framework.
I hope the changes make sense.

### DIFF
--- a/boringNotch/boringNotchApp.swift
+++ b/boringNotch/boringNotchApp.swift
@@ -105,6 +105,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             let windowY = screenFrame.frame.height
 
             window.setFrameTopLeftPoint(NSPoint(x: windowX, y: windowY))
+            let notchSize = setNotchSize()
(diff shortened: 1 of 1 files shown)


richardkunkli:
Hi,
Thanks for the PR, really appreciated. We are rewritten the whole layout and behavior of the notch. Hopefully it will now work as expected in that scenario. Keep your eyes on the updates and let me know if the issue still persist. If you are familiar with Xcode you can build it for yourself using the `panda-glows` branch.

iamharshdev:
@TinyTanic let me know if you can update the pr with panda-glows branch.

richardkunkli:
This won't solve the problem. We should set up an observer for display changes and then run the auto sizing script again. I'll check it by the way tomorrow.

Issue #120, opened 2024-09-25
add haptic feedback for calendar scrolling


freepicheep:
Haptic feedback (on the trackpad) for when users scroll through the calendar view would just be an extra touch.

*I actually have this working on my personal fork and would be happy to open a PR if this is something the team wants.*

![notchCalendar](https://github.com/user-attachments/assets/8d1ac44d-a1c8-4bd8-a93a-9c851862789b)

PR #122, opened 2024-09-25
Haptics for Calendar Scroll 

freepicheep:
Super simple PR for adding haptic trackpad feedback when users scroll through the calendar in boringNotch. It will respect their general preferences for enabling haptics.

I think Xcode reformatted the file, hence all the additions and deletions. I can redo the PR and try to follow the original formatting if you like.

This would close my suggestion on issue #120.


freepicheep:
I quick changed it to use SwiftUI's `sensoryFeedback` instead of importing AppKit and using `NSHapticFeedbackManager`. I tried to follow the convention you used in the `ContentView`.

PR #123, opened 2024-09-27
fixed so tab opens last used or shelf if user said so

freepicheep:
The tab will always open to last view unless you have a file in the tray and have it set to always open tray if a file is in it.

### DIFF
--- a/boringNotch/components/Settings/SettingsView.swift
+++ b/boringNotch/components/Settings/SettingsView.swift
@@ -293,7 +293,6 @@ struct SettingsView: View {
                          """)
                 }
             }
-            .disabled(!vm.alwaysShowTabs)
             Toggle("Enable boring mirror", isOn: $vm.showMirror)
             Picker("Mirror shape", selection: $vm.mirrorShape) {
                 Text("Circle")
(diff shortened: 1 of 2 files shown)


freepicheep:
I just added a new setting to always enable the last tab. It is disabled if `alwaysShowTabs` is disabled because a user could get stuck in the shelf if it is empty and the tabs aren't always shown. 

I am allowing `openLastTabByDefault` even if `openShelfByDefault` is enabled. The preference will always be for `openShelfByDefault` if there is a file in the shelf. If `openShelfByDefault` is enabled and the new `openLastTabByDefault` is enabled, a note informs the user which one receives priority.

freepicheep:
I polished up a few things:
1. The `openLastTabByDefault` is not disabled if `alwaysShowTabs` is disabled; it will just enable `alwaysShowTabs` if toggled on. 
2. I fixed an issue where the user could get stuck in the files shelf by disabling `alwaysShowTabs` when the shelf was empty and set to be the current view.

@richardkunkli, It should be ready to merge now. Here is a demo.

https://github.com/user-attachments/assets/0e1c97e9-2f6b-43b7-a006-4962701953fb

richardkunkli:
Great work! The animation for the header seems off a bit, it's broken for some reason. I'll fix, but the logic is great. 🚀

PR #143, opened 2024-10-20
fix album flickering due to artwork being updated unnecessarily

Alexander5015:
The album art and background color flicker due to being updated unnecessarily, added a condition to check if the new album art is not nil and is not the old art before updating artwork


richardkunkli:
Hi,
Great additions, many thanks !!

Issue #98, opened 2024-09-10
Enh req: Timeline control


rodalpho:
When I'm playing music and bring up the notch media player, I would like a timeline starting at 0:00 and ending at the length of the song. I should be able to click on that timeline to skip around/ahead.

richardkunkli:
Hi,
We have already considered this feature, might include in a later release.

PR #144, opened 2024-10-20
[FEATURE] Add a music slider

Alexander5015:
To address issue #98, add a slider that allows users to view the current song's elapsed time and duration, as well as seek through the track. The slider is designed to match the average color of the album art.

https://github.com/user-attachments/assets/b460d3d6-8cab-4cde-9fec-eff084745100

### DIFF
--- a/boringNotch/components/Notch/NotchHomeView.swift
+++ b/boringNotch/components/Notch/NotchHomeView.swift
@@ -7,13 +7,18 @@
 
 import SwiftUI
 import Defaults
+import Combine
 
 struct NotchHomeView: View {
     @EnvironmentObject var vm: BoringViewModel
     @EnvironmentObject var musicManager: MusicManager
     @EnvironmentObject var batteryModel: BatteryStatusViewModel
     @EnvironmentObject var webcamManager: WebcamManager
-    
+
+    @State private var sliderValue: Double = 0
+    @State private var dragging: Bool = false
+    @State private var timer: AnyCancellable?
+    @State private var previousBundleIdentifier: String = "com.apple.Music"
     let albumArtNamespace: Namespace.ID
     
     var body: some View {
@@ -73,6 +78,15 @@ struct NotchHomeView: View {
                             MarqueeText(musicManager.songTitle, font: .headline, nsFont: .headline, textColor: .white, frameWidth: geo.size.width)
                             MarqueeText(musicManager.artistName, font: .headline, nsFont: .headline, textColor: .gray, frameWidth: geo.size.width)
                                 .fontWeight(.medium)
+
+                            MusicSliderView(sliderValue: $sliderValue,
+                                            duration: musicManager.songDuration,
+                                            color: musicManager.avgColor,
+                                            dragging: $dragging) { newValue in
+                                musicManager.seekTrack(to: newValue)
+                            }
+                            .padding(.top, 5)
+                            .frame(height: 36)
                         }
                     }
                     .padding(.top, 10)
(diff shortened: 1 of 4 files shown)


richardkunkli:
Hi,
Awesome work!! 🔥 I will work on that a bit, but there is just one problem. If you click on the slider it will seek to the right position, but it will lost the duration and state and won't be good until pause and play again.

boringNotch/components/Notch/NotchHomeView.swift
```diff
@@ -73,9 +78,18 @@ struct NotchHomeView: View {
                             MarqueeText(musicManager.songTitle, font: .headline, nsFont: .headline, textColor: .white, frameWidth: geo.size.width)
                             MarqueeText(musicManager.artistName, font: .headline, nsFont: .headline, textColor: .gray, frameWidth: geo.size.width)
                                 .fontWeight(.medium)
+
+                            MusicSliderView(sliderValue: $sliderValue,
+                                            duration: musicManager.songDuration,
+                                            color: musicManager.avgColor,
+                                            dragging: $dragging) { newValue in
+                                musicManager.seekTrack(to: newValue)
```
richardkunkli, line 86: Overwrite sliderValue locally to even smoother experience

boringNotch/components/Notch/NotchHomeView.swift
```diff
@@ -131,6 +145,118 @@ struct NotchHomeView: View {
...
+        GeometryReader { geometry in
+            let width = geometry.size.width
+            let height = geometry.size.height
+            
+            ZStack(alignment: .leading) {
+                // Background track
+                Rectangle()
+                    .fill(Color.gray.opacity(0.3))
+                    .frame(height: height / 3) // Track height
+                    .cornerRadius(height / 6)
```
richardkunkli, line 231: Use Capsule() shape instead of Rectangles and no need for cornerRadius

boringNotch/components/Notch/NotchHomeView.swift
```diff
@@ -131,6 +145,118 @@ struct NotchHomeView: View {
...
+
+                // Filled track
+                Rectangle()
+                    .fill(color)
+                    .frame(width: (CGFloat(value - range.lowerBound) / CGFloat(range.upperBound - range.lowerBound)) * width, height: height / 3)
```
richardkunkli, line 236: I believe a logic like `(range.upperBound - range.lowerBound) == .zero ? 0 : (range.upperBound - range.lowerBound) * (value - range.lowerBound) * width` maybe better, cause there are invalid frame dimension errors

boringNotch/components/Notch/NotchHomeView.swift
```diff
@@ -131,6 +145,118 @@ struct NotchHomeView: View {
...
+        let timeDifference = currentTime.timeIntervalSince(musicManager.timestampDate)
+        // Calculate the real-time elapsed time
+        let currentElapsedTime = musicManager.elapsedTime + (timeDifference * musicManager.playbackRate)
+        sliderValue = min(currentElapsedTime, musicManager.songDuration) // Clamp to song duration
+    }
+}
+
+struct MusicSliderView: View {
+    @Binding var sliderValue: Double
+    var duration: Double
```
richardkunkli, line 181: Make the duration to be `@State`

richardkunkli (commented):
Small changes, in order to be more future-proof and eliminate errors

boringNotch/components/Notch/NotchHomeView.swift
Alexander5015: Should this be `@Binding`?(idk I don't have a lot of swift experience)

boringNotch/components/Notch/NotchHomeView.swift
Alexander5015: I'm not sure what you mean

richardkunkli: sliderValue is dependent on polling. It gets the value from a function that fetches the current playing media. So it has some delay. if you explicitly overwrite it under `musicManager.seekTrack(to: newValue)` line, like `sliderValue = newValue` then it ensures there will be no jumping between values.

boringNotch/components/Notch/NotchHomeView.swift
richardkunkli: You are right, `@Binding`

boringNotch/components/Notch/NotchHomeView.swift
Alexander5015: Oh ok, I'm not sure this would change anything because sliderValue is also bound to value which gets updated as soon as the thumb of the slider changes. When the slider is released, it then runs the seekTrack callback function with newValue == sliderValue. 

The jumping that I tried to fix happened because the poll of the musicManager directly after the seek returns values for timestamp and elapsed time from before the seek takes effect. This resulted in the slider returning back to where it would be if the seek hadn't happened until the seek takes effect and updates the musicManager state, setting the thumb back to where the user dragged it. To fix this I added a small delay after releasing where dragging is true (keeping the sliderValue locked) until I assumed the musicManager had the updated values. However, I could use timestamp to check whether the local or polled value is correct as well, which would probably be the better way to do this. Did you experience jumps when testing it?

richardkunkli: I am experiencing this

# Preview

https://github.com/user-attachments/assets/432cc25e-8748-419e-9b70-d6616682ffed

Alexander5015: Can you try with the latest changes? This seems to be a duration issue but I can't replicate it

richardkunkli: It's working now. Good job. 🚀

Issue #134, opened 2024-10-19
[BUG] Gesture to open notch extremely laggy.


aryansidbatte:
When having hover to open disabled. Using the gesture to open the notch has large animation lag / low framerate

I am on an M1 Max chip Macbook Pro

PR #145, opened 2024-10-20
Fix slow open gesture

Alexander5015:
Improved the gesture handling logic to optimize the open gesture, resolving issue #134.


richardkunkli:
Hi,
Thanks a lot. Much needed changes.

Issue #149, opened 2024-10-20
[BUG]Notchnotopening


ayan-binish:
**Describe the bug**
Boring Notch version Glowing Panda only opens once. It boots up perfectly fine and when I play music. When I use the notch to pause/unpause content, the Boring Notch closes unexpectetly and gives me error code.

**To Reproduce**
1. Go to 'Boring Notch '
2. Click on 'the Notch'
3. Pause/Unpause content

**Expected behavior**
I expected Boring Notch Glowing Panda to work as seamlessy or even better than Sleeping Snail, but I keep on getting the error code.

**Screenshots or recordings**
![image](https://github.com/user-attachments/assets/904554cb-5e59-4977-9c5a-48998c349015)

**Additional context**
Add any other context about the problem here. Mentioning what version are you using.
I am using Boring Notch Glowing Panda with MacOS 15.0.1.
**Checks**
- [x] I haven't found any duplicates with my issue.

iamharshdev:
Can you share some crash logs.

ayan-binish:
I'll try

ayan-binish:
This is all that I found:
Process: boringNotch [1437]
Path: /Applications/boringNotch.app/Contents/MacOS/boringNotch
Identifier: theboringteam.boringnotch
Version: 2.1 (2.1+1)
Code Type: ARM-64 (Native)
Parent Process: launchd 1

User ID: 501
Date/Time: 2024-10-20 12:10:25.4405 -0700
OS Version: macOS 15.0.1 (24A348)
Report Version: 12
Anonymous UUID: 5C2284FE-108B-8FF3-C05A-9FCBC23FFAFD
Time Awake Since Boot: 580 seconds
System Integrity Protection: enabled
Crashed Thread: 0 Dispatch queue: com.apple.main-thread
Exception Type: EXC_BREAKPOINT (SIGTRAP)
Exception Codes: 0x0000000000000001, 0x00000001844611fc
Termination Reason: Namespace SIGNAL, Code 5 Trace/BPT trap: 5
Terminating Process: exc handler [1437]
Application Specific Backtrace 0:
0 CoreFoundation 0x00000001806bcec0 exceptionPreprocess + 176
1 libobjc.A.dylib 0x00000001801a2cd8 objc_exception_throw + 88
2 AVFCapture 0x000000019f916494 -[AVCaptureSession_Tundra stopRunning] + 504
[...]

PR #154, opened 2024-10-21
Add better thread management to WebcamManager

Alexander5015:
Not sure if this would help with #149 but I noticed the UI would freeze while starting up the camera so I think this should at least fix that by moving management of the capture session to a .userInitiated sequential queue so it doesn't use the main thread. I also added in more weak references to self

### DIFF
--- a/boringNotch/managers/WebcamManager.swift
+++ b/boringNotch/managers/WebcamManager.swift
@@ -20,6 +20,7 @@ class WebcamManager: NSObject, ObservableObject {
             objectWillChange.send()
         }
     }
+    private let sessionQueue = DispatchQueue(label: "BoringNotch.WebcamManager.SessionQueue", qos: .userInitiated)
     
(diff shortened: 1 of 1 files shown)


boringNotch/managers/WebcamManager.swift
```diff
@@ -20,54 +20,75 @@ class WebcamManager: NSObject, ObservableObject {
...
-        guard let session = captureSession else { return }
-        
-        session.beginConfiguration()
-        session.sessionPreset = .high
-        
-        guard let videoDevice = AVCaptureDevice.default(.builtInWideAngleCamera, for: .video, position: .unspecified),
-              let videoInput = try? AVCaptureDeviceInput(device: videoDevice)
-        else {
-            return
+        sessionQueue.async { [weak self] in
```
iamharshdev, line 31: Looks good.

PR #160, opened 2024-10-23
fix(calendar): enhance WheelPicker scrolling performance

HXavier21:
While scrolling through the calendar's WheelPicker, I noticed that the selected date would jump directly to the left edge, preventing users from easily selecting dates located on the right side of the list through scrolling alone.

To address this, I made some adjustments by adding spacers to both ends of the list, allowing users to select any date through simple scroll gestures. Additionally, I optimized the dateButton selection logic: whether the user selects a date by scrolling or tapping, the selected date will now automatically center in the view (this part involves some hardcoding, and I welcome feedback for further improvement).

I also refactored the dateButton to minimize unnecessary recalculations, transforming it into a stateless component for better performance and maintainability.

### DIFF
--- a/boringNotch/components/Calender/BoringCalendar.swift
+++ b/boringNotch/components/Calender/BoringCalendar.swift
@@ -32,7 +32,7 @@ struct WheelPicker: View {
             HStack(spacing: config.spacing) {
                 let totalSteps = config.steps * (config.past + config.future)
                 let spacerNum = config.offset
-                ForEach(0..<totalSteps + 2 * spacerNum, id: \.self) { index in
+                ForEach(0..<totalSteps + 2 * spacerNum + 1, id: \.self) { index in
                     if(index < spacerNum || index > totalSteps + spacerNum - 1){
                         Spacer().frame(width: 24, height: 24).id(index)
                     } else {
--- a/boringNotch/components/Calender/BoringCalendar.swift
+++ b/boringNotch/components/Calender/BoringCalendar.swift
@@ -21,27 +21,35 @@ struct WheelPicker: View {
     @Binding var selectedDate: Date
     @State private var scrollPosition: Int?
     @State private var haptics: Bool = false
+    @State private var byClick: Bool = false
     let config: Config
     
     var body: some View {
         ScrollView(.horizontal, showsIndicators: false) {
             HStack(spacing: config.spacing) {
                 let totalSteps = config.steps * config.count
-                ForEach(0..<totalSteps, id: \.self) { index in
-                    dateButton(for: index)
+                ForEach(0..<totalSteps + 6, id: \.self) { index in
+                    if(index < 3 || index > totalSteps + 2){
+                        Spacer().frame(width: 24, height: 24).id(index)
+                    } else {
+                        let date = dateForIndex(index)
+                        let isSelected = isDateSelected(index)
+                        dateButton(date: date, isSelected: isSelected)
+                    }
                 }
             }
             .frame(height: 50)
             .scrollTargetLayout()
         }
         .scrollIndicators(.never)
-        .scrollPosition(id: $scrollPosition)
+        .scrollPosition(id: $scrollPosition,anchor: .top)
         .safeAreaPadding(.horizontal)
         .sensoryFeedback(.alignment, trigger: haptics)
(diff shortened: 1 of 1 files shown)


freepicheep:
Hey, I love the work you did, @HXavier21! That behavior has been annoying me for a little while now. 

I noticed that it has a bug though when you scroll to the beginning of your calendar. It doesn't keep any day selected. Here's a short screen recording showing the problem.

https://github.com/user-attachments/assets/fce477e9-9a2d-4d30-b76f-176da4b63bf4

HXavier21:
> Hey, I love the work you did, @HXavier21! That behavior has been annoying me for a little while now.
> 
> I noticed that it has a bug though when you scroll to the beginning of your calendar. It doesn't keep any day selected. Here's a short screen recording showing the problem.
> 
>  CleanShot.2024-10-23.at.14.49.28.mp4

It seems there are some differences between the Preview, compile-run, and actual device performance. When running with *Boring Mirror* enabled, you might notice it's unusually large, making the calendar display interval very narrow. But I feel like I haven't modified the relevant code? Which makes it difficult for me to confirm what the actual running effect is.
image
I've attempted to fix the issue by tweaking the number of spacers, but that approach seemed not so good. On top of that, I considered another solution where we could pass configuration settings to the `handleScrollChange` function to determine if the selected date has hit its limit. However, I'm not too keen on this method either, as it doesn't seem to promote good code decoupling.

HXavier21:
@freepicheep I refactored the code to allow customization of the number of past and future days in the list, as well as the offset of the selected date within the display range (number of elements reserved to the left). This will facilitate adding relevant settings in the settings interface later.

Additionally, I added restrictions on scrolling, so it is now impossible to slide to non-existent dates, albeit at the cost of a slightly odd feel.If you attempt to continue sliding after selecting the date at either extreme end, it will forcefully snap back to a valid date.

HXavier21:
> @freepicheep I refactored the code to allow customization of the number of past and future days in the list, as well as the offset of the selected date within the display range (number of elements reserved to the left). This will facilitate adding relevant settings in the settings interface later.
> 
> Additionally, I added restrictions on scrolling, so it is now impossible to slide to non-existent dates, albeit at the cost of a slightly odd feel.If you attempt to continue sliding after selecting the date at either extreme end, it will forcefully snap back to a valid date.

If you also find this feel odd, I have an alternative solution:
When you attempt to continue sliding after selecting the date at either extreme end, the list will continue to slide but the selected date will not change. Which do you prefer? @freepicheep 

Here's what it looks like when sliding to the limit:
image

HXavier21:
Identified and fixed issues with the `past` and `future` configurations. Now, user can freely modify `past`, `future`, and `offset` to achieve any desired effect.
Added a spacer on the right to prevent the ***Notch*** from being too long, ensuring that the rightmost end can be scrolled to.

iamharshdev:
> > Hey, I love the work you did, @HXavier21! That behavior has been annoying me for a little while now.
> 
> > 
> 
> > I noticed that it has a bug though when you scroll to the beginning of your calendar. It doesn't keep any day selected. Here's a short screen recording showing the problem.
> 
> > 
> 
> >  CleanShot.2024-10-23.at.14.49.28.mp4
> 
> 
> 
> It seems there are some differences between the Preview, compile-run, and actual device performance. When running with *Boring Mirror* enabled, you might notice it's unusually large, making the calendar display interval very narrow. But I feel like I haven't modified the relevant code? Which makes it difficult for me to confirm what the actual running effect is.
> 
> image
> 
> I've attempted to fix the issue by tweaking the number of spacers, but that approach seemed not so good. On top of that, I considered another solution where we could pass configuration settings to the `handleScrollChange` function to determine if the selected date has hit its limit. However, I'm not too keen on this method either, as it doesn't seem to promote good code decoupling.

To solve this @richardkunkli might need to rewrite the whole layout using cells.

richardkunkli:
Absolutely, the change is much needed to support future features and customizations even more. I'll check this code throughoutly, then will merge it. Great work!

freepicheep:
@HXavier21 very nice work! I tried it and it looks great.🔥

freepicheep:
Thanks for all your hard work on this issue!

richardkunkli:
Definitely better, good job!🚀

PR #164, opened 2024-10-23
Fix Boring Notch window Level and implement Restart Boring Notch

Alexander5015:
Fix Boring Notch sitting under menubar extra windows and implement the Restart Boring Notch menu item


iamharshdev:
Great work @Alexander5015

Issue #163, opened 2024-10-23
[FEATURE] Sneak Peek Keyboard Shortcut


freepicheep:
**Is your feature request related to a problem? Please describe.**
na

**Describe the solution you'd like**
I would like the ability to call the sneak peek using a keyboard shortcut. I do this a lot with MediaMate currently, and it shouldn't be too hard to add to boringNotch. I'd be happy to work on this myself if you assign it to me. I'm opening a issue/feature request to gather input before I get into it too far. Also, closing issues is fun. I would plan on using Sindre's [package](https://github.com/sindresorhus/KeyboardShortcuts). I would create a new tab called "Shortcuts" in the settings.

**Additional context**
Shortcuts are nice.

**Checks**
- [x] I haven't found any duplicates with my issue (I know that #135 is similar, but that is a much broader issue...let's take it in chunks).

freepicheep:
Lol, I just realized we already use Sindre's package.

PR #165, opened 2024-10-23
added keyboard shortcut for sneak peek

freepicheep:
This pr does the following:
1. Creates a new tab in settings called "Shortcuts"
2. Adds a shortcut for users to toggle the sneak peak (for music).

It addresses #163.


iamharshdev:
Looks good

Issue #176, opened 2024-10-29
[BUG] Sneak Peek shortcut not working


bronisMateusz:
**Describe the bug**
Sneak Peak shortcut and record shortcut not working.

**To Reproduce**
1. Go to Settings
2. Click on Shortcuts
3. Try perform a shortcut
4. Try register a new shortcut
5. Perform shortcut didn't show sneak peek and register didn't register a new shortcut

**Expected behavior**
Show sneak peek and register new shortcut.

**Screenshots or recordings**
![Zrzut ekranu 2024-10-29 o 10 02 20](https://github.com/user-attachments/assets/a736f67a-49f5-4302-9674-0967b39b354b)

**Additional context**
- Boring Notch 2.2
- Mac OS 15.1 (24B83)
- MacBook Pro 14 2021

**Checks**
- [x] I haven't found any duplicates with my issue.

freepicheep:
It's working for me on macOS 15.2 on a MacBook Pro 14 2021. I also tried it in a virtual machine with 15.1 and it worked fine. 
https://github.com/user-attachments/assets/9fdb7bfe-9238-4259-8376-f2f4b27bcd0b

Maybe use an app like [Pearcleaner](https://github.com/alienator88/Pearcleaner) to uninstall it then try again.

bronisMateusz:
Ok, I have just checked more thoroughly. And it doesn't work only in combination with function keys, for example `command+F6`.

In addition, if you do manage to set up this shortcut, the sneak peak does not show up anyway. The notch gently dims for a moment, but nothing shows underneath it.

freepicheep:
> Ok, I have just checked more thoroughly. And it doesn't work only in combination with function keys, for example `command+F6`.

Actually, I just checked and it does work with that shortcut if your function keys are set to be standard function keys instead of special features by default. If they are set to the special features (which it sounds like they are), you need to press `⌘ + fn + F6`. I'm still opening a PR to make the shortcut easier, but I wanted to clarify the issue.

freepicheep:
Wonderful! You're very welcome. 

I don't think we can configure it to show the `fn` unfortunately...we rely on Sindre's [Keyboard Shortcuts](https://github.com/sindresorhus/KeyboardShortcuts) library for global shortcut functionality and I don't think it would be worth forking it and putting in the work to add that feature. Richard just merged my PR for setting the default hotkey to `⌘ + Shift + H`, so hopefully it won't be too confusing in the future.

[...]

PR #177, opened 2024-10-29
Easier default shortcut for sneak peek

freepicheep:
The current default keyboard shortcut for the sneak peek is `⌘ + F6`. For most people on a MacBook, their function keys are set to use the special features instead of standard functions, meaning they would need to press `⌘ + fn + F6`. This isn't very intuitive or clear. Technically it works (I tested it), but some users might not understand that. This was part of the issue with #176 and perhaps some others who want to use the shortcut. I'm suggesting that the default be `⌘ + Shift + H` which this simple PR would do.


richardkunkli:
Indeed a better default value.

Issue #173, opened 2024-10-28
[FEATURE] Low power mode color


Illyze:
**Is your feature request related to a problem? Please describe.**
When in power mode the color is still white or green but not orange.
**Describe the solution you'd like**

**Additional context**

**Checks**
- [x] I haven't found any duplicates with my issue.

freepicheep:
That's a great idea. I just made a working version and will open a PR when I polish it up. Hopefully it will be ready for the next release.

PR #180, opened 2024-10-30
Support for low power mode battery visual

freepicheep:
Currently the battery in the notch header does not properly indicate when your Mac is in low power mode (#173). I added support for this by accessing `ProcessInfo.processInfo.isLowPowerModeEnabled` and setting up an Notification observer to update the `BatteryStatusViewModel` when this changes. If the battery is in low power mode, it will show yellow, *regardless of the percentage and charge mode*. I believe this is consistent with how Apple does it.


richardkunkli:
Awesome! Great changes!

PR #184, opened 2024-11-03
Placeholder provided by media player

overflow-sudo:
If the artwork data cannot be found, it will be replaced with a placeholder provided by media player.

Example: When the user uses Zen Browser, the artwork doesn't change, and the previous song's artwork is displayed. However, if the artwork is equal to nil, the code sets the Zen Browser icon as the artwork.


richardkunkli:
Hey,
First of all, thank you for the contribution. Much appreciated. Can you please use the `dev` branch instead of `main`. It won't be difficult to migrate, due to those are synced right now. I review the code until and merge it if the change has been performed.

overflow-sudo:
Hey,
I have changed the base branch to dev.

Thank you.

Issue #135, opened 2024-10-19
[FEATURE] Keyboard shortcuts


5183nischal:
It would be pretty cool to be able to call the notch without moving the mouse to it, for example via a keyboard shortcut! I'm sorry if this is already an option and I missed it but didn't find it in the settings. What do you guys think?

richardkunkli:
Hi,
Thanks for the feedback. It could be useful for keyboard-first users. We will implement it in a future update.

freepicheep:
I've been thinking about this more. So I implemented the ability to trigger the sneak peek with a shortcut with #165. @5183nischal, does this new functionality do what you want, or would you want a shortcut to trigger the entire boringNotch app?

5183nischal:
This is great! And I like the media sneak preview! At the risk of sounding too demanding, what would also be cool is to replicate the toggle behavior of the notch when the mouse it take to it with a keyboard shortcut. I tend to hide my menubar and when I take the mouse to the notch it not only brings the notch (which I want) but also shows the menubar (which I don't want). Would be even cooler to have a shortcut to directly toggle the shelf or home with different keybindings (or move between the two and closing when pressing the same shortcut repeatedly). I don't know how complicated this is but I'd really enjoy it. Thanks!

5183nischal:
This is great -- pretty much exactly what I had in mind! looking forward to release.

PR #191, opened 2024-11-07
Open Entire Notch with Shortcut

freepicheep:
## What it does
Record or use the standard shortcut of `Command + Shift + I` to open the entire notch. The notch will remain open for 3 seconds and then close. The notch will close if you press the shortcut again.

## Why?
Someone requested this feature in #135. I think it is a good idea. Honestly, I'll probably use it a lot more than the sneak peek shortcut because I like to see the album art bigger if I'm trying to figure out what song is playing. 

## Changes
1. I added a new shortcut variable to the `KeyboardShortcuts`
2. Registered this with the `onKeyDown` function in the `AppDelegate`. Inside the `AppeDelegate` I added the logic for dispatching a work item because this will allow me to cancel the notch close dispatch if the user closes the notch before the timeout.
3. A setting in the keyboard shortcuts to change the default shortcut.

I chose to name it "Toggle Notch Open" because it also closes the notch.


freepicheep:
Here is a demo of how it works.

https://github.com/user-attachments/assets/2b4ae7b4-2e64-49ba-9f97-e53fa993f0b6

Alexander5015:
Looks good!

Issue #151, opened 2024-10-21
[FEATURE] Pick and choose specific calendars


tomfoolery03:
**Is your feature request related to a problem? Please describe.**
I have shared calendars with my family where numerous events that I dont care about are shared with me. Boring Notch does not have an option to select specific apple calendars to show.

**Describe the solution you'd like**
An option to select/deselect which calendars show up on the notch

**Additional context**
<img width="163" alt="image" src="https://github.com/user-attachments/assets/3e0367c6-beb2-4f76-835a-6cb645c66a72">
Just wanna express my gratitude for this app, great work!
**Checks**
- [x] I haven't found any duplicates with my issue.

richardkunkli:
Hi,
Thanks for the feedback. Indeed this is a nice feature we should implement.

PR #253, opened 2024-12-19
[FEATURE] Choose which calendars to show events from

AdrianGri:
### Problem
This PR addresses the issue brought up in #151 where the specific calendars to load events from cannot be customized.

### Solution
Add new settings menu to customize which calendars to pull events from.

Changes:
- Create new calendar tab in settings
   - Move show calendar toggle to this tab
   - Add a list of all of the user's calendars with toggles beside them
- Add calendar selection state to constants
  - The state is stored as an enum either all or a set of selected calendar identifiers
  - all is the default behavior where all calendars are used

### Screenshots
Calendar settings menu (deleted emails for privacy)
Calendar menu

### I'd really appreciate any feedback, I'm open to making modifications!


theboringhumane:
We need to use some other intuitive UI for this cc @richardkunkli

AdrianGri:
Hmmm maybe some different UI like the old control center settings page with + / - buttons to add / remove and two distinct lists for select and unselected? @theboringhumane 

image

theboringhumane:
Yeah. This one's better cc @richardkunkli

PR #309, opened 2025-01-16
Exclude finder in fullscreen

christiaansp:
finder was being seen as a fullscreen app so i excluded it from being detected. (btw for the role in the discord i'm Chris (display name) UID is 738324643324100629


Alexander5015:
Closing because I think this has been addressed now

PR #353, opened 2025-02-01
fix: improve notch closing animation

eexxio:
Replace sliding circle animation with a clean vertical transition. The notch now moves straight up when closing, maintaining better visual coverage of the hardware notch throughout the animation.

### DIFF
--- a/boringNotch/ContentView.swift
+++ b/boringNotch/ContentView.swift
@@ -35,6 +35,18 @@ struct ContentView: View {
     @Default(.useMusicVisualizer) var useMusicVisualizer
 
     @Default(.showNotHumanFace) var showNotHumanFace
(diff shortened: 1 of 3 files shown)


theboringhumane:
Looks good, but'll merge it with the changes to give them option to select the animation either have the wobble or don't have it. Thanks @eexxio

eexxio:
should I add a menu for allowing them to choose?

theboringhumane:
Yes let's give them Options.

Mahasvan:
I was looking through the PRs to check if the smooth animations were brought back. Thanks!

I noticed that Sneak Peek is broken, just wanted to let you know!
Clicking cmd+shift+H makes the notch dim for a split second. It also doesn't work when the media changes (which it did in the previous version.)

image

Issue #357, opened 2025-02-02
[FEATURE] enable 0 height option


yaxarat:
**Is your feature request related to a problem? Please describe.**
I want to be able to create a dynamic notch with 0 height. Working on an external screen that does not have a notch, the one created by boring notch quickly becomes an eye sore.

**Describe the solution you'd like**
I'd love a feature where we can set it to 0 height and let it just show on hover.

**Additional context**

**Checks**
- [x] I haven't found any duplicates with my issue.

ardaberktin:
That would actually be great!

PR #359, opened 2025-02-02
feat: Allow 0 height notch

yaxarat:
# In this PR
Introduced the ability to set custom notch height that is less than 10.

Closes: https://github.com/TheBoredTeam/boring.notch/issues/357

## Demo

https://github.com/user-attachments/assets/40da08f0-69a4-46ca-9e22-e4a86eee4d78


yaxarat:
One issue with this change is that when the height is too low (anything below 6), it causes the closing animation to over shoot, leading to a runtime error (**does not** crash) of `Invalid view geometry: height is negative.`. I've tried clamping the height, but no luck.

yaxarat:
@theboringhumane just checking in as this PR has been sitting for close to a month.

theboringhumane:
This will be merged by today hopefully

theboringhumane:
@yaxarat can you rebase it with origin/dev

yaxarat:
@theboringhumane closing this since rebase really messed up the git history. I've put up the same PR here: #397

PR #385, opened 2025-02-19
feat:  improve WheelPicker snapping & Haptic behavior

ShirakawaMio:
### Changes:
1. **Haptic**: Previously, haptic feedback was triggered when clicking the trackpad, which caused a slightly conflicting sense with the system’s native trackpad haptic.
2. **WheelPicker**: Improved alignment behavior to ensure strict snapping to buttons, making the picker movement more stable.

Before: 

https://github.com/user-attachments/assets/6c684167-26e4-4073-9f5e-fda634ee2e77

sometimes failed to snap perfectly to button positions and crolling could drift slightly down and upward

After:

https://github.com/user-attachments/assets/0ef894e4-c82f-4775-9147-c4fa1e5c3ff2


theboringhumane:
@ShirakawaMio rebase with main

PR #477, opened 2025-04-04
New audio model

menaxmaged:
image
New audio player with support to view the output device if airpods


Alexander5015:
Thanks for the contribution! I think this still needs a few changes before we can merge it. First, it should detect built-in speakers from various Apple products and use the appropriate symbols. Additionally, it should detect AirPods Pro and AirPods Max. I also think we might want to redesign this, but I’m not much of a designer, so for now, start with the detection.

menaxmaged:
ok i will work on it.  thank you

Alexander5015:
Awesome, thanks! I am really busy working on the next release and trying to resolve a licensing issue, so it may take me some time to get to this. Also, please note we are changing our license to GPL3.

menaxmaged:
can you tell me more about the new license cause i don't completely understand the differences
btw i am working to use the audio detector  to make like a notification with the device and the battery status 
like when you connect airpods to iphone 
but i don't know which module should i use the system live activtites or the inline hud or make it using Expanding view type there is many options can you suggest me which is better or the most organized

Alexander5015:
Disclaimer, please note that this is not legal advice. I’ve attached GitHub’s summary of the new license’s clauses below for your reference. The old license, CC BY-NC-ND 4.0, is problematic because it’s not intended to be a software license but actually a media license. The “NC” (non-commercial) and “ND” (no derivatives) clauses have changed, allowing commercial use and modifications (which is obviously crucial for open-source software). Also, any derivative of the software must have public source code, preventing someone from taking part or all of our work and using it in a closed-source app.
 
image

For the UI, maybe look at how the battery notifications work, but I didn't work on any of these so I don't really know how they work.

Alexander5015:
Please rebase your branch and resolve conflicts

Alexander5015:
I'm closing this for now since it hasn't been updated. Feel free to open a new PR if you are still interested in adding this

PR #486, opened 2025-04-07
Album Art Blur for Live Activity

legojrp:
I’ve modified the Album Art blurs for the Live Activity Album Art to overlay the Play icon and apply a blur effect. 

It also handles brightness of the album art for darker or brigher album arts. 

For example, 

Screenshot 2025-04-07 at 17 20 28

Becomes this: 
Screenshot 2025-04-07 at 17 16 04

I can't seem to find precedent among Apple apps or OS of album art or images at such a low resolution, but at least on my  Macbook Air M2, the album arts are quite bad and pixelated, and that causes issues of aliasing (I think that's the right word, I'm not a computer graphics person). 

Maybe this could become a setting, as this is more user preference than anything. But I think it looks better.

### DIFF
--- a/boringNotch/ContentView.swift
+++ b/boringNotch/ContentView.swift
@@ -344,9 +344,22 @@ struct ContentView: View {
                 Color.clear
                     .aspectRatio(1, contentMode: .fit)
                     .background(
-                        Image(nsImage: musicManager.albumArt)
-                            .resizable()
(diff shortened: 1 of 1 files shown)


Alexander5015:
I prefer the existing look, closing for now. Perhaps a less blurred version would look better?

PR #489, opened 2025-04-08
Added Weather in Header

legojrp:
Added functionality of weather to be displayed in the header. 
It requires location services, user will be prompted for their location. It also requires an API Key from [OpenWeatherMap.org](https://OpenWeatherMap.org), which is free as long as someone doesn't open their notch 1000 times per hour. 

Showcase: 

https://github.com/user-attachments/assets/8adebb52-4656-4a1c-80da-a95cdae189b3

It's off by default, and won't request your location unless you specifically go in settings to turn it on. 

*Disclaimer: I did use some AI on this. I'm not a great SwiftUI person, so if there is anything wrong or bad syntax or whatever, my apologies.*


Alexander5015:
Thanks for the contribution! I will review it fully after the next release. One thing I noticed when briefly skimming it is that you accidentally added entitlements for Apple Music and Spotify. Not a big deal, but you should remove those so it doesn't ask for Spotify permission before the user selects Spotify in settings.

A few notes for features that should be included(I don't know if these exist or not):
- Ability to use a fixed location rather than location services 
- Save results and only update if the last update was more than 5 or 10 minutes ago, or the location has changed
- Get an updated location using a fixed rate as well

Issue #459, opened 2025-03-28
[FEATURE] Shelf Upgrade


Git-Wolf53:
**Is your feature request related to a problem? Please describe.**

_No._

**Describe the solution you'd like**

_Implement a toggleable setting that automatically deletes files when dragged out of the Shelf.
Another desirable toggleable feature of the Shelf is that once a file is dragged out, the “Notch” automatically returns to its default position in the main area of the “Notch”._

**Additional context**

**Checks**
- [x] I haven't found any duplicates with my issue.

BenWilles:
+1 Would love this too and also the ability to automatically delete the file at the source once it's dropped to another place.

thiemohaller:
I would also love to have files grouped within the shelf (similar to Dropover), e.g. I want to move files from multiple directories and upload them in a "Drop here" file upload on some website move them to a different directory all at once - it would be great if I could just drop the files as I find them in the shelf, then head to the website and drop them all in at once.

gsportelli:
+1 for automatically closing shelf and removing items upon dragging them off the shelf

Alexander5015:
## File Shelf Overhaul (Upcoming Update)

The next update introduces a complete redesign of the file shelf in Boring Notch. This update focuses on faster file management, improved usability, and addressing many requests raised in this and other issues. One notability feature not yet implemented is groups of files in the shelf. but you are able to select and perform actions on multiple files, so this is somewhat addressed in the update.

### Key Changes

#### 1. Redesigned UI
A fully refreshed interface for smoother navigation and improved clarity.

#### 2. Multi-Item Selection
Support for selecting multiple items:
- **Shift + Click** → consecutive selection  
- **Command + Click** → non-consecutive selection

#### 3. New Context Menu
Right-click now provides many different actions directly in the shelf, including:
- **Open**
- **Open With…**

#### 4. Double-Click to Open
Files can now be opened via double-click, matching Finder behavior.

#### 5. Easier File Removal
Files can be removed after dragging by default (no extra steps required).

#### 6. Move vs. Copy Behavior
- Files **move by default** when dragged into the shelf. This will remove the file from its original location after the file is dragged from the notch.
- Hold **Option** to copy.
- Copying can also be set as the default in settings.

Issue #432, opened 2025-03-10
[FEATURE] Ability to open files from the shelf


JayanAXHF:
**Is your feature request related to a problem? Please describe.**
Sometimes, there are files/dirs, that i really open very frequently. However its annoying to navigate back to finder when i need to.

**Describe the solution you'd like**
Ability to click a file or directory in the shelf and for it to either open it in the default or open it in finder. 

**Additional context**
I'm not sure if this is possible or not.

**Checks**
- [x] I haven't found any duplicates with my issue.

Alexander5015:
Either double clicking or right click context menu now allow you to open files in boring notch v2.7

PR #516, opened 2025-04-23
Add open file on tap to shelf

azhao4227:
I know there's a licensing issue preventing the review of PRs, but I thought I'd add this quick one for now :)

As mentioned in #432 and #459, it would be nice to have the ability to open files from the shelf directly. I implemented this as a toggleable feature, as shown below:

https://github.com/user-attachments/assets/1ba90c84-1524-4125-a272-4531fab0702b


Alexander5015:
Thanks for your contribution, @azhao4227, this looks great! I am able to review this now, could you please rebase your branch with the latest changes? Also, I think for now let's make this the default behaviour, we can add in a setting for this later if we need to, but I feel like we already have a bit too many settings :)

Issue #384, opened 2025-02-18
[FEATURE] Reminder/Task List


SaketKulkarni121:
**Is your feature request related to a problem? Please describe.**
I always want to see what the items I have to do are and when they are due, but I hate opening the reminders app on macOS because its
really annoying to use.

**Describe the solution you'd like**
In the notch, allow for a setting that looks at your reminders, with the same implementation as the calendar. When enabled, the user can slide up and down with 2 fingers on the month of the calendar and it will get replaced with the reminders. The reminders will have a top bar with the reminder list name and just the list of things. These can be pulled directly from Apple without any changes. From here, the user can scroll through, and check off any reminder completed. I think it makes more sense to force the user to input new reminders on the app itself because doing that through the notch seems quite difficult. 

**Checks**
- [x] I haven't found any duplicates with my issue.

PR #520, opened 2025-04-25
Initial Implementation of Reminders Widget

azhao4227:
I decided to implement a reminders widget per issue #384 since I also would like to see which reminders I have coming up as a quick glance. This functions very similarly to the calendar widget, and once the user would like to mark the reminder as completed, they can just click on the reminder and it will disappear with an animation (see below demonstration).

NOTE: I had to increase the width of open notch window to get this to fit in (esp. when all four widgets are present). Would love to get some feedback on this!

https://github.com/user-attachments/assets/be4b8343-865a-4dfe-8d3c-7723fc88de6e


SaleemTheKing:
I would love to see this feature implemented as soon as possible, this is what im really missing in the app.

Also i have a question about the reminders. Are reminders with a set time/date also visible from the calendar widget?

Alexander5015:
Please update this by rebasing with `dev` and resolving merge conflicts

Alexander5015:
Implemented in the new calendar

Issue #495, opened 2025-04-12
License Issue


Alexander5015:
Hey @Oscillix @HXavier21 @urmit-chauhan @ShirakawaMio @eexxio @Kraigo  @overflow-sudo @andreamoretti890

Please read this, it is very important. I need to change our license for the repository, because for some reason we currently use CC BY-NC-ND 4.0, which is not designed to be a software license. I encourage you to do your own research about this but one of the key problems with it is that ND stands for no derivatives, which means that contributors would be in violation of this license. To change it I need consent from every contributor so I am reaching out here to get your consent. I want to use GPL 3 as our new license, and I have attached the GitHub summary of this license below. 

Thanks!

![Image](https://github.com/user-attachments/assets/22a7684d-9a6a-4363-a613-dccdbdfcece9)

eexxio:
sounds good👍

ShirakawaMio:
Great😺

HXavier21:
It's okay🥳

Oscillix:
Seems good to me 👍

bronisMateusz:
@Alexander5015  Is this still a blocker for the official release of the new version of the app?

Alexander5015:
Yes, but it should be resolved soon

PR #529, opened 2025-04-26
Added Multi Media Controller and Shuffle/Repeat

pranav1st:
Implemented 2 enhancements:

- Modified the existing MediaControllers to make a unified Media Controller that can be switched between each other for Apple Music, Spotify and Youtube Music.
- Added the Shuffle Toggle for each MediaController.

https://github.com/user-attachments/assets/073a7b7b-8277-4b0f-9b8d-118811e0868f

Kindly share feedback or reviews if necessary, since this is my first PR :)
Thank you.


Alexander5015:
Thanks for your contribution, @pranav1st, this looks great! I’m currently unable to accept PRs due to a licensing issue (see https://github.com/TheBoredTeam/boring.notch/issues/495 for more information), but I’ll review it once it’s resolved. One quick comment though: Shuffle should go on the other side. Shuffle is usually on the left, and repeat is on the right.

pranav1st:
No problem, hope it gets resolved soon. Thanks for the feedback, I've made the corrections as specified, and added the Repeat Toggle feature for the controllers.

Just a few points to note regarding the Repeat toggle:
| Music Controller   | Repeat Modes Supported          | Icon Behavior |
|----------------|---------------------------------|--------------|
| Apple Music   | Repeat Off, Repeat All, Repeat One | Icon color and Icon changes accordingly |
| Spotify       | Repeat Off, Repeat All | Icon color and Icon changes accordingly |
| YouTube Music | Repeat Off, Repeat All, Repeat One | Icon is static |

Spotify Controller does not support Repeat One since I couldn't find any direct way to implement Repeat One.
Youtube Music Controller has static color icon since it requires continuous API calls to track its state which might lead to increased resource utilization

Alexander5015:
Perfect! Here’s one more piece of feedback: we should be able to get the icon for YT Music. Currently, we’re assuming that certain aspects of the playback state, like the seek position (other than regular playback), won’t change while the notch is open. This is because the notch is usually accessed via the mouse, and the series of events that would need to occur for this to be an issue isn’t realistic. So, I would add a state check to the force update, which is called when the notch is opened.

pranav1st:
Hey! Apologies for the late reply, haven't been able to work lately on it due to other commitments. Just getting back to working on this now.

I haven't been able to reproduce the issue where the playback state won’t change while the notch is open, but I've added the code for force update nevertheless, although I don't see any changes in behavior. Please do let me know on how I could reproduce this issue?

Sharing my discord username if you prefer discussing there or for further clarifications - pranav.m1 | Nickname - Zenetix (already joined the server)

Alexander5015:
Hi Pranav, I’m really sorry about the delay in getting back to you about the PR. 

I’ve just resolved the licensing issue and want to add this feature to the next update. I won’t be able to take a full look at it until next week, but I was saying that since we have a state check when the notch opens, we can manually poll YouTube for its shuffle and repeat state. This should prevent any inconsistencies (if you click the button, the next time you open the notch, it will poll and get the updated state).

pranav1st:
Hi, glad to hear the licensing issue has been resolved! I've made a couple of updates to the PR:
- Added a forced playback state polling option that triggers both when the notch opens as mentioned and when the repeat/shuffle button is clicked.
- Fixed a bug where the seek-to action wasn't being registered by the API, which prevented it from functioning as expected.

Alexander5015:
This PR should be split into two: one for shuffle/repeat and another for the multi-media controller. Reviewing each feature separately will make it much easier and clearer.

Alexander5015:
shuffle/repeat have been added so I will close this now. Open a new PR for multiple media controllers, and I will review it.

PR #543, opened 2025-05-10
Create SECURITY.md

anouar-bm:


freepicheep:
So, what is this?

Ein-Tim:
That's not really helpful as it is a PR only providing the template of the SECURITY.md file. I guess it would be better if one of the maintainers would actually create a SECURITY.md file.

Alexander5015:
Closing, see #674

PR #578, opened 2025-06-05
Add default keyboard shortcut for toggling the notch

Anipaleja:
This PR introduces a default keyboard shortcut (⌘ + ⌥ + N) for toggling the notch view.
Previously, the shortcut was undefined until set manually in the settings. This improves out-of-the-box usability.

Changes:

- Added .toggleNotch name under KeyboardShortcuts.Name extension
- Set a default shortcut (⌘ + ⌥ + N) only if none exists
- Avoided use of the non-existent KeyboardShortcuts.defaults property for compatibility
This change is backwards-compatible and will not override existing user-defined shortcuts.


Abdullah-03:
Low quality PRs harm open source projects. If you don't know how the code base works, AI will be no good as demonstrated by your file changes.

PR #579, opened 2025-06-06
Music Gestures

mateocabanal:
This PR implements horizontal gestures to skip to the next/previous songs. 

I used the previous `.panGesture` view modifier, which required a minor modificiation.
The callback type had to be changed from `() -> Void` to `() -> Bool?`, where if the returned value is `true`, the `PanGesture` handler resets the accumulated scroll deltas. Returning `nil` or `false` will skip the reset.

This was used to avoid retriggering events, which in this case was skipping multiple songs.

https://github.com/user-attachments/assets/0344bb77-05f7-4edd-8abf-d4fbd301cbf1


Alexander5015:
please set the base branch to `dev`

Alexander5015:
I'm closing this since it is stale, feel free to open a new PR. If you do, make sure you use `dev` as the base branch and not `main`(which is what GitHub defaults to)

PR #646, opened 2025-07-29
Fix UI freezes by replacing blocking semaphores with async/await in music controllers

weitheng:
Media controllers were using patterns that could freeze the UI. `SpotifyController` and `AppleMusicController` used semaphores for `AppleScript` calls, while `YouTubeMusicController` used `DispatchQueue` and completion handlers.

Changes:
- Replaced semaphore-based blocking with async/await in AppleMusic/Spotify controllers
- Converted YouTubeMusicController from DispatchQueue/callbacks to async/await
- Added @preconcurrency imports for Swift 6 readiness
- Updated to modern Task.sleep(for:) syntax
- Added async protocol methods where missing

This fix prevents UI freezes during:
- App initialization when music apps are running
- Playback state change notifications
- User interactions (seek, shuffle, repeat)                                                                                                                                                                

The async implementation allows the main thread to remain responsive while AppleScript execution happens in the background. All controllers now follow consistent async patterns and are ready for Swift 6 strict concurrency.

### DIFF
--- a/boringNotch/MediaControllers/AppleMusicController.swift
+++ b/boringNotch/MediaControllers/AppleMusicController.swift
@@ -20,9 +20,9 @@ class AppleMusicController: MediaControllerProtocol {
     // MARK: - Initialization
     init() {
         setupPlaybackStateChangeObserver()
-        DispatchQueue.main.async { [weak self] in
-            if self?.isActive() == true {
-                self?.updatePlaybackInfo()
+        Task { @MainActor in
+            if isActive() {
+                await updatePlaybackInfoAsync()
             }
         }
     }
@@ -30,7 +30,7 @@ class AppleMusicController: MediaControllerProtocol {
     private func setupPlaybackStateChangeObserver() {
         DistributedNotificationCenter.default().addObserver(
             self,
-            selector: #selector(updatePlaybackInfo),
+            selector: #selector(handlePlaybackStateChange),
             name: NSNotification.Name("com.apple.Music.playerInfo"),
             object: nil
         )
@@ -68,7 +68,9 @@ class AppleMusicController: MediaControllerProtocol {
     
     func seek(to time: Double) {
         executeCommand("set player position to \(time)")
-        updatePlaybackInfo()
+        Task { @MainActor in
+            await updatePlaybackInfoAsync()
+        }
     }
     
     func toggleShuffle() {
@@ -101,8 +103,22 @@ class AppleMusicController: MediaControllerProtocol {
     }
     
     // MARK: - Private Methods
-    @objc func updatePlaybackInfo() {
-        guard let descriptor = fetchPlaybackInfo() else { return }
+    @objc private func handlePlaybackStateChange() {
+        Task { @MainActor in
+            await updatePlaybackInfoAsync()
+        }
+    }
+    
+    // Public method for protocol conformance
+    func updatePlaybackInfo() {
+        Task { @MainActor in
+            await updatePlaybackInfoAsync()
+        }
+    }
+    
+    @MainActor
+    private func updatePlaybackInfoAsync() async {
+        guard let descriptor = try? await fetchPlaybackInfoAsync() else { return }
         guard descriptor.numberOfItems >= 8 else { return }
         
         let isPlaying = descriptor.atIndex(1)?.booleanValue ?? false
@@ -131,9 +147,7 @@ class AppleMusicController: MediaControllerProtocol {
             artwork: (artworkData?.isEmpty ?? true) ? nil : artworkData
         )
         
-        DispatchQueue.main.async { [weak self] in
-            self?.playbackState = updatedState
-        }
+        self.playbackState = updatedState
     }
     
     private func executeCommand(_ command: String) {
@@ -143,7 +157,7 @@ class AppleMusicController: MediaControllerProtocol {
         }
     }
     
-    private func fetchPlaybackInfo() -> NSAppleEventDescriptor? {
+    private func fetchPlaybackInfoAsync() async throws -> NSAppleEventDescriptor? {
         let script = """
         tell application "Music"
             set isRunning to true
@@ -175,13 +189,7 @@ class AppleMusicController: MediaControllerProtocol {
             end try
         end tell
         """
-        var descriptor: NSAppleEventDescriptor? = nil
-        let semaphore = DispatchSemaphore(value: 0)
-        Task {
-            descriptor = try? await AppleScriptHelper.execute(script)
-            semaphore.signal()
-        }
-        semaphore.wait()
-        return descriptor
+        
+        return try await AppleScriptHelper.execute(script)
     }
 }
--- a/boringNotch/MediaControllers/AppleMusicController.swift
+++ b/boringNotch/MediaControllers/AppleMusicController.swift
@@ -16,7 +16,7 @@ class AppleMusicController: MediaControllerProtocol {
     )
     
     var playbackStatePublisher: Published<PlaybackState>.Publisher { $playbackState }
-    private var notificationObserver: Any?
+    private var notificationTask: Task<Void, Never>?
     
     // MARK: - Initialization
     init() {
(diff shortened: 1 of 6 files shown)


boringNotch/MediaControllers/SpotifyController.swift
```diff
@@ -67,17 +67,23 @@ class SpotifyController: MediaControllerProtocol {
     
     func seek(to time: Double) {
```
Alexander5015: You should probably just make executeCommand and all these playback control functions async

boringNotch/MediaControllers/SpotifyController.swift
```diff
@@ -114,27 +134,23 @@ class SpotifyController: MediaControllerProtocol {
...
         
-        DispatchQueue.main.async { [weak self] in
-            self?.playbackState = state
-        }
+        self.playbackState = state
         
         // Load artwork asynchronously and update the state when complete
         if !artworkURL.isEmpty, let url = URL(string: artworkURL) {
-            DispatchQueue.global(qos: .background).async { [weak self] in
+            Task {
```
Alexander5015: This is on the main thread, move to background

boringNotch/MediaControllers/AppleMusicController.swift
```diff
@@ -20,17 +20,17 @@ class AppleMusicController: MediaControllerProtocol {
...
+                await updatePlaybackInfoAsync()
             }
         }
     }
     
     private func setupPlaybackStateChangeObserver() {
         DistributedNotificationCenter.default().addObserver(
             self,
-            selector: #selector(updatePlaybackInfo),
+            selector: #selector(handlePlaybackStateChange),
```
Alexander5015: Just eliminate the selector here and call the Task directly

weitheng: DistributedNotificationCenter requires a selector method, so we can't directly use a Task

Alexander5015: For async there is DistributedNotificationCenter.default.notifications
https://developer.apple.com/documentation/foundation/notificationcenter/notifications(named:object:)

But it works with a block as well:
```
 DistributedNotificationCenter.default().addObserver(
    forName: Notification.Name("com.apple.Music.playerInfo"),
    object: nil,
    queue: .main) { notification in
   
    }
```

boringNotch/MediaControllers/SpotifyController.swift
Alexander5015: By this I meant making play, pause, etc. all async functions in MusicManager, the protocol and in the controllers

boringNotch/MediaControllers/SpotifyController.swift
```diff
@@ -92,20 +102,20 @@ class SpotifyController: MediaControllerProtocol {
     }
     
     // MARK: - Private Methods
+    // DistributedNotificationCenter requires a selector method, so we can't directly use a Task
```
Alexander5015: See my reply about other ways to register a notification observer

boringNotch/MediaControllers/SpotifyController.swift
```diff
@@ -114,38 +144,34 @@ class SpotifyController: MediaControllerProtocol {
...
-        DispatchQueue.main.async { [weak self] in
-            self?.playbackState = state
+        await MainActor.run {
+            self.playbackState = state
         }
         
         // Load artwork asynchronously and update the state when complete
         if !artworkURL.isEmpty, let url = URL(string: artworkURL) {
-            DispatchQueue.global(qos: .background).async { [weak self] in
+            Task.detached {
```
Alexander5015: This should probably still use a weak self reference

boringNotch/MediaControllers/AppleMusicController.swift
```diff
@@ -101,8 +121,22 @@ class AppleMusicController: MediaControllerProtocol {
     }
     
     // MARK: - Private Methods
-    @objc func updatePlaybackInfo() {
-        guard let descriptor = fetchPlaybackInfo() else { return }
+    // DistributedNotificationCenter requires a selector method, so we can't directly use a Task
```
Alexander5015: See my reply about other ways to register a notification observer

boringNotch/MediaControllers/AppleMusicController.swift
weitheng: Ahh yes, thanks for the reference!

boringNotch/MediaControllers/YouTubeMusicController.swift
```diff
@@ -391,62 +389,56 @@ class YouTubeMusicController: MediaControllerProtocol {
...
+            updatePlaybackInfo()
+            
+        } catch {
+            if let urlError = error as? URLError, urlError.code == .userAuthenticationRequired {
+                // Authentication error - token might be expired
+                accessToken = nil
+                authenticateAndSetup()
+                
+                // Try the command again after authentication
+                try? await Task.sleep(nanoseconds: 2_000_000_000) // 2 seconds
```
Alexander5015:
```suggestion
                try? await Task.sleep(for: .seconds(2))
```

boringNotch/MediaControllers/YouTubeMusicController.swift
```diff
@@ -148,36 +148,32 @@ class YouTubeMusicController: MediaControllerProtocol {
...
-                        self?.updatePlaybackInfo()
+                
+                if httpResponse.statusCode == 401 || httpResponse.statusCode == 403 {
+                    throw URLError(.userAuthenticationRequired)
+                } else if httpResponse.statusCode < 200 || httpResponse.statusCode >= 300 {
+                    throw URLError(.badServerResponse)
+                }
+                
+                // Update playback info after seeking
+                try? await Task.sleep(nanoseconds: 300_000_000) // 0.3 seconds
```
Alexander5015: I find the either of these to be cleaner syntax:

```suggestion
                try? await Task.sleep(for: .seconds(0.3))
```
or
```suggestion
                try? await Task.sleep(for: .milliseconds(300))
```

boringNotch/MediaControllers/YouTubeMusicController.swift
```diff
@@ -186,25 +182,27 @@ class YouTubeMusicController: MediaControllerProtocol {
...
     }
     
-    func toggleShuffle() {
-        sendCommand(endpoint: "/shuffle", method: "POST")
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) { [weak self] in
-            self?.fetchShuffleState()
-        }
+    func toggleShuffle() async {
+        await sendCommand(endpoint: "/shuffle", method: "POST")
+        try? await Task.sleep(nanoseconds: 200_000_000) // 0.2 seconds
```
Alexander5015:
```suggestion 
        try? await Task.sleep(for: .seconds(0.2))
```
or
```suggestion 
        try? await Task.sleep(for: .milliseconds(200))
```

boringNotch/MediaControllers/YouTubeMusicController.swift
```diff
@@ -186,25 +182,27 @@ class YouTubeMusicController: MediaControllerProtocol {
...
     }
 
-    func toggleRepeat() {
-        sendCommand(endpoint: "/switch-repeat", method: "POST")
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) { [weak self] in
-            self?.fetchRepeatMode()
-        }
+    func toggleRepeat() async {
+        await sendCommand(endpoint: "/switch-repeat", method: "POST")
+        try? await Task.sleep(nanoseconds: 200_000_000) // 0.2 seconds
```
Alexander5015:
```suggestion 
        try? await Task.sleep(for: .seconds(0.2))
```
or
```suggestion 
        try? await Task.sleep(for: .milliseconds(200))
```

Alexander5015 (changes requested):
I left some suggested changes for cleaner syntax.

Also, YoutubeMusicController needs to be updated further to prevent mixing synchronous and asynchronous code. You should also use this: https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/migrationstrategy to enable warnings for  all the media controller files to ensure they use proper concurrency.

PR #707, opened 2025-08-24
Add idle auto-collapse for notch + General settings (delay & pointer tolerance)

azocca99:
## Summary
This PR adds an **idle auto-collapse** feature for the island/notch and exposes two user-configurable controls in **Settings → General → Notch behavior**:
- **Auto-collapse on idle** (toggle)
- **Idle delay** (0.5–5.0 s)
- **Pointer tolerance** (1–8 px, filters tiny trackpad drift)

The goal is to let the island close itself after a short period of pointer stillness, while staying friendly to trackpads (jitter tolerance) and preserving the existing open-on-hover feel.

---

### New
- `IdlePointerManager.swift`
  - Tracks global mouse movement, posts `Notification.Name.bnPointerDidGoIdle` after `idleInterval` if movement ≤ `movementTolerance`.
  - API: `start()`, `arm()`, `stop()`; configurable `idleInterval`, `movementTolerance`, `isEnabled`.

### Updated
- `BoringViewCoordinator.swift`
  - Owns an `IdlePointerManager` instance.
  - Arms idle on expand / sneak-peek open.
  - `applyIdleSettings()` reads `@AppStorage` and applies settings live.
  - Posts/handles idle to collapse programmatically.

- `ContentView.swift`
  - Listens for `.bnPointerDidGoIdle` and closes the notch with animation.
  - Small hover smoothing so moving the pointer into the island doesn’t instantly close it.

- `Settings/SettingsView.swift` (General → Notch behavior)
  - Added UI:
    - **Auto-collapse on idle** (toggle)
    - **Idle delay** slider (single-row, label-in-slider) — **0.5…5.0 s** in **0.1 s** steps
    - **Pointer tolerance** slider (single-row) — **1…8 px**
  - Live-applies settings via `BoringViewCoordinator.shared.applyIdleSettings()`.

- Project cleanup
  - Removed duplicate `IdlePointerManager 2.swift` and consolidated to `managers/IdlePointerManager.swift`.

## Defaults / Behavior
- Default values currently ship as:
  - `idleEnabled = true`
  - `idleSeconds = 5.0`
  - `idleTolerancePx = 2`
- Happy to flip `idleEnabled` default to **false** if you prefer opt-in to preserve exact legacy behavior on first launch.

---

## How to use
1. Open **Settings → General → Notch behavior**.
2. Toggle **Auto-collapse on idle**.
3. Adjust **Idle delay** (e.g., 2–3 s feels good).
4. If the notch never collapses on trackpads, increase **Pointer tolerance** (6–8 px usually works well).

---

## QA / Test Plan
- [ ] With feature **enabled**, open the island, keep pointer still → island auto-closes after the selected delay.
- [ ] With feature **disabled**, behavior is identical to current main.
- [ ] Tiny pointer wiggles below tolerance do **not** reset the timer; larger moves do.
- [ ] Hover-open flow still works; user can move into the island to click buttons without instant collapse.
- [ ] Settings changes apply live (no relaunch required).
- [ ] App runs from a standalone build (Release) without Xcode.

---

## Implementation notes
- Idle event is delivered via `Notification.Name.bnPointerDidGoIdle` (posted on the main runloop).
- `IdlePointerManager` installs a global & local monitor (mouse moved/dragged) and resets its timer on significant movement.
- Hover logic received a small debounce/grace so interacting with content doesn’t immediately close the island.

---

## Risk / Compatibility
- Low. Feature is behind a user toggle via `@AppStorage`, and the “close” path reuses existing coordinator APIs.
- No additional entitlements/permissions needed.
- Localization: only uses existing strings; can add keys if you prefer localized labels.

---

## Follow-ups (optional)
- Consider exposing **Open animation speed** as a slider (currently `.bouncy.speed(4.0)`).
- If you prefer opt-in by default, I can switch `idleEnabled` default to **false** in this PR.

---

Thanks for reviewing! Happy to adjust naming, defaults, or UI styling to match project conventions.


Alexander5015:
Could you set the target branch to `dev`?

boringNotch/components/Settings/SettingsView.swift
```diff
@@ -144,6 +138,10 @@ struct GeneralSettings: View {
     @Default(.automaticallySwitchDisplay) var automaticallySwitchDisplay
     @Default(.enableGestures) var enableGestures
     @Default(.openNotchOnHover) var openNotchOnHover
+    // Idle auto-collapse settings
+    @AppStorage("idleEnabled") private var idleEnabled: Bool = true
```
Alexander5015, line 142: These should be defined as Defaults in the Constants file

boringNotch/IdlePointerManager.swift
```diff
@@ -0,0 +1,99 @@
+#if os(macOS)
```
Alexander5015, line 1: This isn't necessary, it is guaranteed to be running on macOS

Alexander5015 (changes requested):
There are two files for the IdlePointerManager, please choose one implementation and delete the other file.

Issue #695, opened 2025-08-14
[FEATURE] Note taking feature


x8vf:
It would be really useful to have a note taking feature similar to the one that is in notchnook

PR #732, opened 2025-09-01
Add Notes Tab

omeriadon:
#### This PR introduces a new “Notes” tab with the following features:

- Toggle monospace font per note.
- Search bar to filter notes.
- Multiple notes sorted by most recently edited.
- Stored in Application Support.

I also made some adjustements to Settings to make it macOS 26 compatible, but it is still backwards compatible with macOS 15 and earlier.

SCR-20250901-rwii-2


omeriadon:
Resolves #695

Alexander5015:
Thanks for the PR! Could you isolate your commits to just the changes for this PR so it doesn't include commits from main?

omeriadon:
Done

boringNotch/components/Notch/NotchNotesView.swift
```diff
@@ -0,0 +1,184 @@
...
+//  NotchNotesView.swift
+//  boringNotch
+//
+//  Created by Adon Omeri on 31/8/2025.
+//
+
+import SwiftUI
+
+struct NotchNotesView: View {
+	@StateObject var notesManager = NotesManager()
```
Alexander5015: I don't think this should be created here, it should be injected or managed elsewhere at a higher level to avoid recreating it every time the view is re-initialized

boringNotch/managers/NotesManager.swift
Alexander5015, line 1: File I/O must not run on the main thread. Change all reads/writes to Swift concurrency

boringNotch/components/Notch/BoringNotchWindow.swift
```diff
@@ -1,50 +1,97 @@
...
+	//
+	//    override var canBecomeKey: Bool {
+	//        false
+	//    }
+	//
+	//    override var canBecomeMain: Bool {
+	//        false
+	//    }
+	// }
+
```
Alexander5015: No need to have this

boringNotch/Localizable.xcstrings
Alexander5015, line 1: I think you may have accidentally removed localization for several strings as they are now marked as `"extractionState" : "stale"`

boringNotch/components/Tabs/TabSelectionView.swift
```diff
@@ -1,55 +1,56 @@
...
+	let icon: String
+	let view: NotchViews
 }
 
 let tabs = [
-    TabModel(label: "Home", icon: "house.fill", view: .home),
-    TabModel(label: "Shelf", icon: "tray.fill", view: .shelf)
+	TabModel(label: "Home", icon: "house.fill", view: .home),
+	TabModel(label: "Shelf", icon: "tray.fill", view: .shelf),
+	TabModel(label: "Notes", icon: "text.pad.header", view: .notes)
```
Alexander5015: This icon won't work on older operating systems, use the deprecated name `note.text`

boringNotch/components/Notch/BoringNotchWindow.swift
```diff
@@ -1,50 +1,97 @@
...
+			.ignoresCycle,
+		]
+
+		isReleasedWhenClosed = false
+		level = .mainMenu + 3
+		hasShadow = false
+	}
+
+		/// Allow this window to become key so it can receive keyboard input
+	override var canBecomeKey: Bool {
```
Alexander5015, line 45: This needs to be fixed so that it does not absorb keyboard input when it is not supposed to

Alexander5015 (changes requested):

omeriadon:
I think I finished all the issues you pointed out, if theres there's anything else like optionally showing that tab please tell me!
Or a button in the bottom right to keep boringNotch open even if the cursor leaves the area

boringNotch/Localizable.xcstrings
omeriadon: When I reverted it to what the repo currently has it still had some errors

boringNotch/components/Tabs/TabSelectionView.swift
Alexander5015: This still needs to be changed I think

boringNotch/animations/FastFadeTransition.swift
```diff
@@ -0,0 +1,18 @@
...
+	//  boringNotch
+	//
+	//  Created by Adon Omeri on 31/8/2025.
+	//
+
+import SwiftUI
+
+extension View {
+	func fastFadeTransition() -> some View {
+		self.transition(.asymmetric(
```
Alexander5015: I don't think you need asymmetric here since both are the same

boringNotch/Localizable.xcstrings
Alexander5015: Right now when I view the localization file, I don't see any of the translations that are currently on `dev`. Also keep in mind that the string extraction is automatically done by Xcode, so the issue with strings being stale is probably in Settings and not the localization itself.

boringNotch/ContentView.swift
```diff
@@ -283,9 +285,23 @@ struct ContentView: View {
                         NotchHomeView(albumArtNamespace: albumArtNamespace)
                     case .shelf:
                         NotchShelfView()
+					case .notes:
+						NotchNotesView().environmentObject(notesManager)
+
                     }
                 }
             }
+			.onChange(of: coordinator.currentView) {
```
Alexander5015, line 302: I'm still having problems with incorrect focus. I haven't fully tested it yet, but one issue is that calling `NSApp.hide(nil)` is problematic since it can also hide the settings window.

Alexander5015 (changes requested):

boringNotch/ContentView.swift
omeriadon: So do i have to add a third variable taht tracks if settings is open to consider?

This would be quite complicated because I think if you prioritise keeping settings open then the notes tab can't give focus back to foreground app

Unless you want me to not make it do that

boringNotch/boringNotchApp.swift
```diff
@@ -122,6 +126,19 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
         )
 
+		window.notesManager = notesManager
+
+		NotesManager.shared.$editorCanFocus
+			.sink { canFocus in
+				if canFocus {
+					self.window?.makeKeyAndOrderFront(nil)
+				} else {
+					self.window?.resignKey()
```
Alexander5015: Why do you call this?

boringNotch/ContentView.swift
Alexander5015: Not necessarily, but the settings window should still behave as expected. You do need a single, centralized focus controller that deterministically decides keyboard focus and ensures it will not absorb keyboard input when the notch is not actively being used. Right now it is split between different files. Because this is not a typical macOS window, it may take some work to find the right way to enable this functionality without causing problems.

omeriadon: Alright I'll try make it, but it will have to be inside AppDelegate not by itself

omeriadon:
I have no idea why the toolbar item has centered itself, other than that the pre-macOS 15 window is good


But the macOS 26 one is broken, I do not know why since it's a very standard `NavigationSplitView` and `Settings { }`

SCR-20250903-scis

Alexander5015:
I am closing this because of the localization issues and it is currently out of date with `dev` Feel free to open a new PR if you want.

Issue #743, opened 2025-09-02
[Bug] All days of the calendar display no events today


M7T5M3P:
### What happened? What are the steps to reproduce the issue?

When mouse hover a day in the calendar functionality it displays 'No event today'

In my example we are the 2nd of september

<img width="198" height="131" alt="Image" src="https://github.com/user-attachments/assets/4c8b62d6-4fd6-4142-bf0c-e1c25044667f" />

### What did you expect to happen?

It would be better to have written 'No events if not today'

### BoringNotch Version

Select a version

### macOS Version

15.3.2

### Relevant log output

```shell

```

### Checks

- [x] I haven't found any duplicates with my issue.

M7T5M3P:
fix this bug here check out my PR : 

[https://github.com/TheBoredTeam/boring.notch/pull/748](url)

PR #748, opened 2025-09-02
Fix calendar localization issues

M7T5M3P:
This PR addresses one of the issues I listed in the main repository, which was quite simple.

Issue https://github.com/TheBoredTeam/boring.notch/issues/743: 

I’ve added the ability to distinguish between "no events today" and "no events on any other day of the year."
Capture d’écran 2025-09-02 à 23 02 00
Capture d’écran 2025-09-02 à 23 02 10

### DIFF
--- a/boringNotch/components/Calendar/BoringCalendar.swift
+++ b/boringNotch/components/Calendar/BoringCalendar.swift
@@ -253,7 +253,7 @@ struct EmptyEventsView: View {
             Image(systemName: "calendar.badge.checkmark")
                 .font(.title)
                 .foregroundColor(Color(white: 0.65))
-            Text(Calendar.current.isDateInToday(selectedDate) ? NSLocalizedString("No events today", comment: "") : NSLocalizedString("No events", comment: ""))
+            Text(Calendar.current.isDateInToday(selectedDate) ? "No events today" : "No events")
                 .font(.subheadline)
                 .foregroundColor(.white)
             Text("Enjoy your free time!")


boringNotch/Localizable.xcstrings
```diff
@@ -13213,6 +13213,106 @@
         }
       }
     },
+    "No events" : {
```
Alexander5015, line 12465: We prefer localization strings only be added to English and translations managed through Crowdin. I know this isn't properly documented yet. The localization its new so I am still working on the details of how it should work.

boringNotch/components/Calendar/BoringCalendar.swift
```diff
@@ -246,12 +246,14 @@ struct CalendarView: View {
...
 struct EmptyEventsView: View {
+    let selectedDate: Date
+    
     var body: some View {
         VStack {
             Image(systemName: "calendar.badge.checkmark")
                 .font(.title)
                 .foregroundColor(Color(white: 0.65))
-            Text("No events today")
+            Text(Calendar.current.isDateInToday(selectedDate) ? NSLocalizedString("No events today", comment: "") : NSLocalizedString("No events", comment: ""))
```
Alexander5015: Refactoring this into a computed property would be more readable imo. something like:
```swift
private var emptyMessage: String {
    if Calendar.current.isDateInToday(selectedDate) {
         return NSLocalizedString("No events today", comment: "")
     } else {
         return NSLocalizedString("No events", comment: "")
    }
}
```

Alexander5015: I also think it would be cool if the text was more dynamic, with more options that can follow "no events" like yesterday, tomorrow, and maybe the date or the day of the week if further in the future or past. We could also have different text for the second line since it's mostly relevant to today as well, which could be a new line for past events and a new line for future events.

Let me know what you think or if you have any suggestions about what this would look like.

boringNotch/Localizable.xcstrings
M7T5M3P: Okay I'll change that !

boringNotch/components/Calendar/BoringCalendar.swift
M7T5M3P: You're sure, I think a ternary is easy to read. But I can also change it if you want

M7T5M3P: Yeah true, I did find my PR empty lol. I was thinking about letting the user create it's own event. Like if there is no event we could have a little + button where he could add one.

For displaying the date or day on further  ones in the week it would be twice the same informations displayed in a small space which is not really recommended in UI/UX.

If you really want to fill this space, we could have an animation (have you ever thought about adding pets to it ?)
Or we could have a fun fact about a date using this api (http://numbersapi.com/2/29/date) which gives fact about what happened on a specific day in the past.

Alexander5015: > For displaying the date or day on further ones in the week it would be twice the same informations displayed in a small space which is not really recommended in UI/UX.

I think just no events yesterday and tomorrow at the very least could work. I agree it is repetitive for further dates which isn't great, but I'm not sure how else to match the existing format.(today is also something that is displayed twice) imo if it is just today that says today and the others are just no events, I would just prefer no events because it is more standardized. 

For the second line, a fun fact could be nice! I'll take a look at the API. something simpler would be to just have a more retrospective text for past days.

Alexander5015: Seems like text from the api might be too long to fix in the small space.

M7T5M3P: You've got a point ! But it's different because it's a special message on today even if it's says it's today. Like if every day say it's their days then it's common to all days. But if it's just today it may do the difference, and it makes the user "feel close to the calendar" (don't really know how to express this feeling lol)

But if you want do avoid that if there is nothing we could have a message like "Nothing planned yet" and no more "today" word.

If it's too long, then we can just have a blank. If you look at apple design (it's in french but it just has one line, no events). So removing this second could also be a solution, give more space. Could be a great idea also, let me know ! 

Capture d’écran 2025-09-03 à 03 51 24

Alexander5015: Just choose what you think is best, maybe try a few options out to see how they look. I don't have a strong preference for any particular direction. If I really don't like it, we can always discuss further or make improvements at some later point. Most of the options we discussed seem better to me than the current text.

Alexander5015:
Since this hasn't been updated in a while, I am going to modify it a bit and then merge it. Feel free to open a new PR if you are interested in adding any of the more interesting variations we discussed.

PR #751, opened 2025-09-03
Feat: animated face, mouse traker, and sleepy eyes

M7T5M3P:
I've added a mouse tracker for the Face display, and a sleepy animation (when the mouse hasn't move for 10 seconds).

Here are the videos.
[Enregistrement de l’écran 2025-09-03 à 03.29.23.zip](https://github.com/user-attachments/files/22106585/Enregistrement.de.l.ecran.2025-09-03.a.03.29.23.zip)

https://github.com/user-attachments/assets/5a434d66-2fb7-4737-842d-4b63dd42dfdf

I had to zip the first video it's too big to upload, it's just me demonstrating the mouse tracking


boringNotch/components/AnimatedFace.swift
```diff
@@ -57,6 +70,120 @@ struct MinimalFaceFeatures: View {
             }
         }
     }
+    
+    func startMouseTracking() {
+        timer = Timer.scheduledTimer(withTimeInterval: 1/60.0, repeats: true) { _ in
```
Alexander5015, line 75: This will be very inefficient. I am pretty sure there are event monitors that would be more performant, but I don't know off the top of my head.

boringNotch/components/AnimatedFace.swift
```diff
@@ -57,6 +70,120 @@ struct MinimalFaceFeatures: View {
...
+        
+        if isSleeping {
+            return CGSize(width: -2, height: eyeSize * 0.5 + 2)
+        }
+        
+        let screenWidth = NSScreen.main?.frame.width ?? 1920
+        let screenHeight = NSScreen.main?.frame.height ?? 1080
+        let screenCenter = CGPoint(x: screenWidth / 2, y: screenHeight / 2)
+        
+        let dx = mouseLocation.x - screenCenter.x
```
Alexander5015, line 161: Should this be based on the pupil location rather than the center of the screen?

boringNotch/components/AnimatedFace.swift
```diff
@@ -66,17 +193,82 @@ struct Eye: View {
...
+        }
+        .onAppear {
+            startZZZAnimation()
+        }
+    }
+    
+    private func startZZZAnimation() {
+        let duration = 3.0
+        
+        Timer.scheduledTimer(withTimeInterval: 1.5, repeats: true) { _ in
```
Alexander5015, line 237: I think it might be possible to do this without timers? Not sure, but make sure to invalidate timers if they are necessary.

boringNotch/components/AnimatedFace.swift
Alexander5015, line 1: I think generally it would be best practice to do the mouse tracking/sleep tracking logic in a separate file.

Alexander5015 (commented):
This looks great! I left some feedback, I will look at it more when I get the chance.

M7T5M3P:
Is this feature still going to be keep ?

Alexander5015:
I left some feedback before which I don't think has been addressed yet. Also, there are other contributors who want to work on new options when idle, so whichever one I merge first has to allow for other options to be selected.

Alexander5015 (changes requested):
Marking as "Changes Requested" to track the outstanding comments from my previous review.

Alexander5015:
closing because stale, but feel free to open a new PR if you are still working on this

PR #757, opened 2025-09-05
Allow blur background

omeriadon:
Uses a gradient between `Material.ultraThinMaterial` and `Color.black`

I tried my best to make it work seamlessly with Music Live Activity and the minised state of Boring Notch, tell me if there are some problems / bugs

In doing this I had to change some componenets like `Calendar` and `BoringHeader` so that their background fade away to opacity 0 instead of black, but same otherwise

<br>

macOS 26:
SCR-20250905-quzi

macOS 15:
SCR-20250905-qvgz

<br>

SCR-20250905-qwhm
SCR-20250905-qwdy
SCR-20250905-qwln


omeriadon:
I'm not sure why its saying I have all those commits before mine because I did rebase it not to have them
I think those are commits that pushed after I started the branch I'm not very good with git

Alexander5015:
you need to change the target branch to dev

omeriadon:
alright

Alexander5015:
Closing, for now we will focus on full black as well as Apple's new liquid glass design.

PR #759, opened 2025-09-06
feat: implemented option for hiding all day events in the calendar.

bueckerlars:
Added an option for hiding all day events form the calendar view using the existing isAllDay property


Alexander5015 (approved):
Looks good!

PR #804, opened 2025-10-20
Feature/screenrecording liveactivity

Ebullioscopic:
# Screen Recording Live Acitivity

https://github.com/user-attachments/assets/69dfeb3d-d27b-4520-aa38-c6ba3d7489fc

I replicated the screen recording live activity functionality from iOS' Dynamic Island

I tested with QuickTime, OBS, Google Meet, MS Teams and Zoom screenshare - it is working on all of them

I have used the ScreenWatcher Private API for making this - it is event driven - and does not do any kind of polling

Lemme know if you need any changes

### DIFF
--- a/boringNotch/managers/ScreenRecordingManager.swift
+++ b/boringNotch/managers/ScreenRecordingManager.swift
@@ -33,7 +33,7 @@ private func screenCaptureEventCallback(eventType: Int32, _: Int32, _: Int32, co
     guard let context = context else { return }
     let manager = Unmanaged<ScreenRecordingManager>.fromOpaque(context).takeUnretainedValue()
     
-    DispatchQueue.main.async {
+    Task { @MainActor in
         print("ScreenRecordingManager: 📢 Screen capture event received (type: \(eventType))")
         manager.checkRecordingStatus()
     }
@@ -55,11 +55,8 @@ class ScreenRecordingManager: ObservableObject {
     
     // MARK: - Private Properties
     private var recordingStartTime: Date?
-    private var durationTimer: Timer?
+    private var durationTask: Task<Void, Never>?
     private var debounceIdleTask: Task<Void, Never>?
-    
-    // MARK: - Configuration
-    private let debounceDelay: TimeInterval = 0.2 // Debounce rapid changes
     
     // MARK: - Initialization
     private init() {
@@ -70,7 +67,7 @@ class ScreenRecordingManager: ObservableObject {
         // Clean up monitoring state
         // Note: We can't call async methods in deinit, so we just clean up local state
         debounceIdleTask?.cancel()
-        durationTimer?.invalidate()
+        durationTask?.cancel()
     }
     
     // MARK: - Public Methods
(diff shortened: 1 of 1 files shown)


Alexander5015:
Could you cherry-pick your commit to keep the git history clean?
It seems like a fairly straightforward change, so the review should be quick, but I’m pretty busy this week so I won't get to it right away.

Alexander5015:
Also, how did you prioritize this compared to the other views?

Ebullioscopic:
Screen recording live activity is a fairly common feature to people using iPhones > 15 and when I released the first iteration of changes to DynamicIsland - people liked that one specifically as it feels more native

Also adding this did not require removing the App Sandbox or adding any other dependencies

That's why I chose this feature to be contributed first as this does not feel very out of place as compared to the Stats one which non developers or non tech people will rarely use

So yeah that's the reason I picked this thing first

Alexander5015:
Oh I meant like if different events are happening including screen recording, how is it prioritized?

Ebullioscopic:
Oh my bad, I didn't prioritize it - just tried turning on the music visualizer after screen recording and it hides the screen recording indicator 

Should I place it at the topmost priority so that it doesn't get hidden?

boringNotch/managers/ScreenRecordingManager.swift
```diff
@@ -0,0 +1,265 @@
...
+            }
+        }
+    }
+    
+    /// Start tracking recording duration
+    private func startDurationTracking() {
+        recordingStartTime = Date()
+        recordingDuration = 0
+        
+        durationTimer = Timer.scheduledTimer(withTimeInterval: 0.1, repeats: true) { [weak self] _ in
```
Alexander5015: This might be better as a single task that sleeps, rather than a repeating timer that creates new tasks each time. The timer here should also probably be slower, 0.1s is probably unnecessary, and something like 0.3-0.5 would be better I think

boringNotch/managers/ScreenRecordingManager.swift
```diff
@@ -0,0 +1,265 @@
...
+    }
+    
+    /// Stop tracking recording duration
+    private func stopDurationTracking() {
+        durationTimer?.invalidate()
+        durationTimer = nil
+        recordingStartTime = nil
+        
+        // Keep the last duration for a moment before resetting
+        DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) { [weak self] in
```
Alexander5015: Use Swift concurrency instead

boringNotch/managers/ScreenRecordingManager.swift
```diff
@@ -0,0 +1,265 @@
...
+    _ context: UnsafeMutableRawPointer?
+) -> Bool
+
+// MARK: - Global Callback Function
+// C function pointer cannot capture context, so we need a global function
+private func screenCaptureEventCallback(eventType: Int32, _: Int32, _: Int32, context: UnsafeMutableRawPointer?) {
+    guard let context = context else { return }
+    let manager = Unmanaged<ScreenRecordingManager>.fromOpaque(context).takeUnretainedValue()
+    
+    DispatchQueue.main.async {
```
Alexander5015: Use Swift Concurrency, it also seems like this will lead to multiple main thread hops

boringNotch/managers/ScreenRecordingManager.swift
```diff
@@ -0,0 +1,265 @@
...
+    @Published var isRecorderIdle: Bool = true
+    @Published var lastUpdated: Date = .distantPast
+    
+    // MARK: - Private Properties
+    private var recordingStartTime: Date?
+    private var durationTimer: Timer?
+    private var debounceIdleTask: Task<Void, Never>?
+    
+    // MARK: - Configuration
+    private let debounceDelay: TimeInterval = 0.2 // Debounce rapid changes
```
Alexander5015: Seems like this variable is unused, you should remove it unless it is necessary

Alexander5015 (commented):
Left some feedback, feature looks good, just needs some tweaking. I’ll make another pass with more feedback once those changes are made.

PR #822, opened 2025-11-12
UI Improvements

Alexander5015:
### Dependency Integration

* Added the `SkyLightWindow` Swift package as a dependency, for lock screen visibility.

### Animation Consistency

* Refactored animation logic in `ContentView.swift` to use a shared `animationSpring` for interactive movements and 
* Removed `.drawingGroup()` from `NotchShape` rendering to potentially improve performance and visual fidelity.

### UI/UX Improvements

* Increased the shadow opacity for the notch when open or hovering, making the UI more visually distinct.
* Added smooth animation for `gestureProgress` changes to enhance transition effects.

### Updated Settings
* Added the option to hide the notch from screen recording
* Added advanced settings tab, to clean up existing settings tabs and reintroduce accent color. 

### Localization Support

* Added multiple new keys to `Localizable.xcstrings` for new UI features


rosmur:
> Increased the shadow opacity for the notch when open or hovering, making the UI more visually distinct.

The shadow size and opacity still feels too light. Alcove's is a good reference. 

Would it be possible to provide us the setting/option to control the shadow size and opacity? or if that is not feasible, requesting an increase in the size and opacity

PR #826, opened 2025-11-14
fix: Typo/grammatical issue in settings toggle label

Precontation:
This is such a small change that I'm not sure if this is considered rude or something in open-source (I'm pretty new to open-source), but just when testing the 2.7-rc.3, I noticed this small thing and just decided to propose this.


Alexander5015:
Not at all, I really appreciate the PR! Small changes are perfect. What I don’t like are big, complex changes that are trying to fix multiple issues at once as it just makes the review more challenging. These should be split up into multiple PRs to make the review process easier and allow me to review and approve each separately.

PR #828, opened 2025-11-14
Add auto-remove shelf items feature

Decryptu:
Implemented an optional setting to automatically remove items from the shelf after they're dragged out to another location.

Screenshot 2025-11-14 at 09 55 30

Changes:
- Added autoRemoveShelfItems setting key in Constants.swift (default: false)
- Added UI toggle in Shelf settings panel
- Created ShelfItemDragSource.swift with file monitoring service
- Updated ShelfItemView.swift to integrate auto-remove on drag completion
- Auto-removal only triggers after successful file copy, handles macOS file conflict dialogs
- Added to Xcode project build target

The feature is opt-in and maintains backward compatibility.

### DIFF
--- a/boringNotch/BoringViewCoordinator.swift
+++ b/boringNotch/BoringViewCoordinator.swift
@@ -241,10 +241,11 @@ class BoringViewCoordinator: ObservableObject {
             if expandingView.show {
                 expandingViewTask?.cancel()
                 let duration: TimeInterval = (expandingView.type == .download ? 2 : 3)
+                let currentType = expandingView.type
                 expandingViewTask = Task { [weak self] in
                     try? await Task.sleep(for: .seconds(duration))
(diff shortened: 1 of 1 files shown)


Alexander5015:
This seems like it might be over-engineered. Is the ShelfItemDragMonitor necessary when the drag delegate callback is available?

Alexander5015:
You might want to use a snippet like this in the callback instead. Let me know if I’m mistaken though!

```swift
 if Defaults[.autoRemoveShelfItems] && operation != .none {
        for item in draggedItems {
            ShelfStateViewModel.shared.remove(item)
        }
    }
```

Decryptu:
Thank you, simplified it

Alexander5015:
Awesome, merged!

Issue #821, opened 2025-11-12
[FEATURE] Automatically scroll calendar to the next event


lambegraham:
**Is your feature request related to a problem? Please describe.**
When I open the notch and have multiple events I have to scroll down to get to the next one.

**Describe the solution you'd like**
It would be good if the notch automatically showed the next upcoming (or current event) when you open it, and then you can scroll up / down for past / future events.

**Additional context**

<img width="642" height="180" alt="Image" src="https://github.com/user-attachments/assets/3a2d2a89-267c-4e40-a874-af6f3b183fe8" />

**Checks**
- [x] I haven't found any duplicates with my issue.

PR #846, opened 2025-11-19
Auto-scroll calendar event list to current/next event

lambegraham:
**Summary**
- Adds automatic scrolling to the calendar event list so the current (in-progress) or next (upcoming) event is positioned at the top of the list when the calendar appears or when events update.

**What changed**
- Updated `EventListView` in BoringCalendar.swift:
  - Replaced the previous `List` usage with a `ScrollViewReader` wrapper and added `.id(event.id)` to each row so rows can be targeted for programmatic scrolling.
  - Implemented `scrollToRelevantEvent(proxy:)` which:
    - Chooses the first event whose `end > now` (in-progress or upcoming).
    - If none, falls back to the last event in the filtered list.
    - Scrolls to the target using the event `id` as the scroll target.
    - Runs the scroll on the main actor with a non-animated transaction to ensure layout is ready.
    - Uses `.top` anchor so the target row appears at the top of the visible scroll area.
  - Triggers the scrolling on `.onAppear` and `.onChange(of: filteredEvents)`.
  - Honors existing filters and settings (`Defaults[.hideCompletedReminders]`, `Defaults[.hideAllDayEvents]`, and the local `@Default(.autoScrollToNextEvent)` flag where present).

**Behavior**
- When the calendar event list appears or the event list changes:
  - If there is an in-progress or upcoming event for the selected date, the list scrolls so that event is at the top of the visible list.
  - If all events are in the past, the list scrolls to the last event.
  - The scroll happens on the main thread after layout to avoid partial offsets or misalignment.
- Each event row is identified by `event.id`, so scrolling is robust across dynamic data updates.

**Files modified**
- BoringCalendar.swift
  - `EventListView`:
    - Added `ScrollViewReader` usage and `.id(event.id)` on rows.
    - Added `scrollToRelevantEvent(proxy:)` and wiring (`.onAppear`, `.onChange`).

**Why**
- Improves usability by surfacing the most relevant event (current or next) immediately, reducing manual scrolling and making the calendar glanceable.

**Edge cases & notes**
- All-day events are filtered when `Defaults[.hideAllDayEvents]` is enabled; the auto-scroll will not target hidden rows.
- If a visual header or other view sits above the list and you still see an offset, we intentionally scroll to `.top`. If needed, the scroll target can be tuned (small spacer above list, slight delay, or switching to a custom `ScrollView` + offset) to achieve pixel-perfect alignment in the containing UI.
- The feature respects the `@Default(.autoScrollToNextEvent)` preference (if enabled in the code) so it can be toggled.

### DIFF
--- a/boringNotch.xcodeproj/project.pbxproj
+++ b/boringNotch.xcodeproj/project.pbxproj
@@ -20,6 +20,11 @@
 		1113ABCE2E80E27000EC13B2 /* ShelfActionService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1113ABBC2E80E27000EC13B2 /* ShelfActionService.swift */; };
 		1113ABD02E80E6BB00EC13B2 /* ThumbnailService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1113ABCF2E80E6BB00EC13B2 /* ThumbnailService.swift */; };
 		111BE95D2ECD71E10079DD4E /* AsyncXPCConnection in Frameworks */ = {isa = PBXBuildFile; productRef = 111BE95C2ECD71E10079DD4E /* AsyncXPCConnection */; };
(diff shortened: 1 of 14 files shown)


lambegraham:
Closes #821

PR #895, opened 2025-12-04
feat: Focus-based HUD disable

bloxyj:
# Updated: Focus-based HUD disable
What changed
Replaced the original "display disconnected" approach with a keyboard focus-based approach. Now the HUD automatically shows on whichever screen has keyboard focus, matching native macOS behavior.

## How it works
When you press volume/brightness keys:

If keyboard focus is on the same screen as the notch → boring.notch HUD shows
If keyboard focus is on a different screen → system HUD shows on that screen

### DIFF
--- a/BoringNotchXPCHelper/BoringNotchXPCHelper.swift
+++ b/BoringNotchXPCHelper/BoringNotchXPCHelper.swift
@@ -138,6 +138,25 @@ class BoringNotchXPCHelper: NSObject, BoringNotchXPCHelperProtocol {
         }
         reply(false)
     }
+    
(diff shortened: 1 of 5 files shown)


Alexander5015:
I think it makes more sense to disable the HUD based on which screen has keyboard focus rather than just if the display is connected or not. Let me know what you think, but this is how I was going to implement it.

Alexander5015:
That way when you are using it on an external display, if you have the notch on your built in display only, you can still see the HUD

bloxyj:
Thanks for pointing that out I can try to update this tomorrow. It’s true that the native macOS HUD behaves like that, so that approach makes more sense.

Issue #894, opened 2025-12-04
[FEATURE] Sneak Peek time delay in Settings


evgeniykl:
1. If possible, add a title and artist name delay to the settings, because it is very fast and very difficult to read everything if the song title is large.

2. It will be great to click on the name of the song in the program itself and go to the program that is currently playing, so as not to search for it among other open programs 

Thank you.

PR #897, opened 2025-12-06
Feature: sneak peak duration

adam-dziuk:
Hi,

Users now can change sneak peak duration in the settings view.

I've added:

- `sneakPeakDuration` as project constant
- UI element in settings view, that shows only if style of SP is set to default
- used that constant when SP animation is called

If you've got any further suggestions about the code (styling/functionality), let me know.

_This PR addresses the first point of [issue #894](https://github.com/TheBoredTeam/boring.notch/issues/894)._

### DIFF
--- a/boringNotch/components/Settings/SettingsView.swift
+++ b/boringNotch/components/Settings/SettingsView.swift
@@ -598,6 +598,7 @@ struct Media: View {
     @Default(.hideNotchOption) var hideNotchOption
     @Default(.enableSneakPeek) private var enableSneakPeek
     @Default(.sneakPeekStyles) var sneakPeekStyles
+    @Default(.sneakPeakDuration) var sneakPeakDuration
 
     @Default(.enableLyrics) var enableLyrics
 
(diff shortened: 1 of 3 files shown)


Alexander5015:
I see no reason not to make this duration setting work for both the inine and standard styles. Also, would it be possible to include an option for the time to be dynamic, allowing for exactly one full loop? This seems to be what most people want when they want longer times. If not that's fine, this is just something I wanted to add in when I was looking at improving sneak peek timing.

adam-dziuk:
Haven't thought about it like that. You're right, it really would be ideal to implement exactly one full loop for the sneak peek feature.

On it.

adam-dziuk:
@Alexander5015 what if we didn't completely ditch the option for users to set that time manually?

I've come up with a helper that calculates the time needed to achieve exactly what we've talked about above - personally, I find it annoying with longer titles.

I think the best approach would be to create the following in settings:

- duration: auto
- duration: fixed

What do you say?

Alexander5015:
> @Alexander5015 what if we didn't completely ditch the option for users to set that time manually?

Yeah this is what I meant about adding the auto option, I think we can have both options.

> I've come up with a helper that calculates the time needed to achieve exactly what we've talked about above - personally, I find it annoying with longer titles.

I was also thinking we could add a maximum duration option to the auto setting, and set this as the default setup with a sufficiently high default max. If we do that, the fixed duration might actually be unnecessary.

Alexander5015:
Closing since this is stale. Feel free to reopen if you are still want this feature.

PR #907, opened 2025-12-12
Fix album art aspect ratio to prevent distortion

sorawitt:
### Changes
Fixed an issue where album art wasn't maintaining its aspect ratio properly, causing non-square artwork to appear stretched or distorted.

### The Issue
The album art display was using `.clipped()` and `.aspectRatio(1, contentMode: .fit)` which didn't handle non-square images well. This resulted in distorted album artwork in both the live activity bar and the expanded view.

### What I Changed

**ContentView.swift (MusicLiveActivity)**
- Changed from `.clipped()` to `.scaledToFill()` before setting frame
- This ensures the image fills the frame while maintaining aspect ratio

**NotchHomeView.swift (albumArtImage)**
- Wrapped the image in a `GeometryReader` to get proper dimensions
- Switched to `.scaledToFill()` with explicit frame sizing
- The image now crops properly instead of stretching

Both changes preserve the existing `matchedGeometryEffect` animations, so transitions still look smooth.

### Before / After

**Before:**
Screenshot 2568-12-12 at 22 47 14
Screenshot 2568-12-12 at 22 47 18

**After:**
Screenshot 2568-12-12 at 22 46 25
Screenshot 2568-12-12 at 22 46 29

Tested with a few different album covers to make sure everything looks correct.

### DIFF
--- a/boringNotch/ContentView.swift
+++ b/boringNotch/ContentView.swift
@@ -388,16 +388,13 @@ struct ContentView: View {
         HStack {
             Image(nsImage: musicManager.albumArt)
                 .resizable()
-                .clipped()
-                .clipShape(
-                    RoundedRectangle(
-                        cornerRadius: MusicPlayerImageSizes.cornerRadiusInset.closed)
-                )
-                .matchedGeometryEffect(id: "albumArt", in: albumArtNamespace)
+                .scaledToFill()
                 .frame(
                     width: max(0, vm.effectiveClosedNotchHeight - 12),
                     height: max(0, vm.effectiveClosedNotchHeight - 12)
                 )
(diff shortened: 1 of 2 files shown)


martinezpl (approved):
Nice, works for me and is definitely an improvement!

daemonsloth:
If thumbnails have text, wouldn't this crop that as well?

Also, for thumbnails where the aspect ratio is more rectangle (i.e. for YouTube videos), wouldn't this crop be too aggressive?

Considering that, I've tried a different approach to preserve the aspect ratio as opposed to cropping the image to fit in the fixed aspect ratio in #932

PR #908, opened 2025-12-13
Fix Apple Music favorite setting syntax in setFavorite method

karadoganyi:
Fixed incorrect syntax in the setFavorite method that caused Apple Music favorites to not work. Updated the request format to match Apple Music API expectations.


Alexander5015:
Please rebase to ensure only your commit is included, without changes from `main`.

PR #919, opened 2025-12-15
Normalize scroll direction for natural scrolling in ScrollMonitor

### LINKED ISSUE #918: [Bug] Gesture direction is reversed when natural scrolling is disabled

I have natural scrolling disabled (in Settings > Trackpad > Scroll & Zoom).
This causes the gesture direction to be reversed (a two finger swipe up opens the notch), which is understandable since the gesture is based on scrolling but doesn't make sense for this app - swiping down should open the notch either way.
Thanks for this great app!

Swiping down should open the notch, swiping up should close it.

### Boring Notch Version

v2.7.3

### macOS Version

15.7.2

None

_No response_

_No response_

### Relevant log output

```shell

```

### Checks

karadoganyi:
**What happened / Steps to reproduce**
With Natural Scrolling disabled in System Settings → Trackpad → Scroll & Zoom, the gesture direction in the app becomes reversed (two-finger swipe up opens the notch), which is unintuitive — users expect the same physical gesture to always open the notch.
**What did you expect to happen?**
Swiping down should open the notch and swiping up should close it — regardless of the system’s scroll preference.
**Solution**
Updated gesture handling logic to normalize scrolling delta values with isDirectionInvertedFromDevice so that gesture direction is based on physical swipe direction, not the system’s natural scroll preference.
This ensures intuitive, consistent gesture behavior whether natural scrolling is ON or OFF.
**Related issue**
Fixes #918(https://github.com/TheBoredTeam/boring.notch/issues/918)

### DIFF
--- a/boringNotch/extensions/PanGesture.swift
+++ b/boringNotch/extensions/PanGesture.swift
@@ -127,9 +127,16 @@ private struct ScrollMonitor: NSViewRepresentable {
             let isAxisDominant: Bool = direction.isHorizontal ? (absDX >= axisDominanceFactor * absDY) : (absDY >= axisDominanceFactor * absDX)
             guard isAxisDominant else { return }
 
+            // Normalize to the physical scroll direction so macOS “Natural scrolling”
+            // does not invert the gesture semantics.
+            let deviceDirectionMultiplier: CGFloat = event.isDirectionInvertedFromDevice ? -1 : 1
+            
             // Scale non-precise (mouse wheel) scrolling deltas so they feel similar to
             // trackpad gestures.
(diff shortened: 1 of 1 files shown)


Alexander5015:
I'm not sure about this. As someone who uses natural scrolling, I'm not sure what the expected behaviour would be for most users of the inverted scrolling. I'm also not sure how this should translate to mouse users using a scroll wheel.

Alexander5015:
I am thinking this could maybe be an option in our advanced settings, but I would want to choose a good default behaviour first, and then have a setting that lets you invert the gesture direction from the norm.

PR #924, opened 2025-12-16
Feature to reduce height of notch for non-notch screens

eltopu:
Added a feature to reduce height of the notch when it is inactive for non-notch displays. Also height adjusts automatically when there is live activity or not

feat: implement dynamic notch height based on live activity state

- Add `effectiveClosedNotchHeight` computed property to adjust notch size
  based on music playback state
- Introduce `inactiveNotchSize` for smaller notch when no media is playing
- Implement fullscreen detection to hide notch when apps are fullscreen
- Add reactive height adjustments that respond to MusicManager playback state
- Optimize notch visibility with `hideOnClosed` property combining fullscreen
  status and user preferences
- Update chin height calculation to account for inactive state
- Improve screen-specific notch rendering with UUID-based tracking

This creates a more dynamic and context-aware notch that conserves screen
space when idle while expanding for active media controls.

### DIFF
--- a/boringNotch/Localizable.xcstrings
+++ b/boringNotch/Localizable.xcstrings
@@ -11708,6 +11708,12 @@
         }
       }
     },
+    "Inactive notch height" : {
+
+    },
+    "Inactive notch sizing (For non-notch displays)" : {
+
+    },
     "Inline" : {
       "localizations" : {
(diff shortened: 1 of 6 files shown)


Alexander5015:
Addressed in #1008

PR #932, opened 2025-12-27
Preserve thumbnail aspect ratios to prevent distortion

daemonsloth:
This change updates the thumbnail rendering to respect each image’s natural aspect ratio, preventing stretching and distortion. Previously, the view enforced a 1:1 aspect ratio using .aspectRatio(1, contentMode: .fit), which caused non-square images to appear squished or letterboxed. By switching to .aspectRatio(contentMode: .fit), the layout now preserves the intrinsic aspect ratio of thumbnails while still fitting them within the available space.

Before the change:
Screenshot 2025-12-28 at 02 34 26

After the change:
Screenshot 2025-12-28 at 02 34 08

### DIFF
--- a/boringNotch/components/Notch/NotchHomeView.swift
+++ b/boringNotch/components/Notch/NotchHomeView.swift
@@ -48,7 +48,7 @@ struct AlbumArtView: View {
                         ? MusicPlayerImageSizes.cornerRadiusInset.opened
                         : MusicPlayerImageSizes.cornerRadiusInset.closed)
             )
-            .aspectRatio(1, contentMode: .fit)
+            .aspectRatio(contentMode: .fit)
             .scaleEffect(x: 1.3, y: 1.4)
             .rotationEffect(.degrees(92))
             .blur(radius: 40)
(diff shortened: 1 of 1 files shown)


Alexander5015:
This is already fixed in another PR

daemonsloth:
> This is already fixed in another PR

Hi @Alexander5015 
I've gone through the PR (#907), and it looks like it uses a different approach to fix this, and that would cause Youtube thumbnails (with 16:9 ratios) to be cropped to a square image losing parts of the thumbnail image.

On the other hand with changes done here, I have tried to preserve the aspect ratio of thumbnail images, hence no crop involved.

I believe this comes down to design choice, may I know your opinion on this?

Alexander5015:
I was actually referring to my PR (I think it’s #887).
This is a screenshot of a YouTube thumbnail from that version:

image

I prefer keeping the width constrained, as shown in the screenshot. I hadn’t fully decided whether to use the cropped square, which I think aligns better with the design before the regression, or stick with my current approach. That’s why the other PR is still open, though I am leaning towards my current implementation.

daemonsloth:
Oh, got it, that looks better. I think a constrained width would make more sense without cropping, since it would better align with macOS’s native Now Playing to maintain visual consistency

PR #937, opened 2025-12-29
Feat: Support Netease Music lyrics and improve LRC parsing

LZWcc:
### 🚀 Description

This PR enhances the accuracy of lyric fetching and provides deep integration for **Netease Music**, a major music service in China.

1. **New Netease Music Source**:
   - Implemented `fetchLyricsFromNetease` to prioritize searching and matching via Netease's API.
   - Fetches official lyrics directly, significantly improving match rates and lyric quality compared to generic web scraping.
2. **Enhanced LRC Parsing**:
   - **The Issue**: The original parser only supported the `[mm:ss.xx]` (2-digit centiseconds) format.
   - **The Fix**: Netease often returns `[mm:ss.xxx]` (3-digit milliseconds). I refactored `parseLRC` using a more robust regex to dynamically handle both 2-digit and 3-digit precision.
3. **Robust Fallback**:
   - If the Netease search fails or returns empty results, the system automatically falls back to `fetchLyricsFromWeb` to ensure reliability.

------

### 📊 Comparison

| **Feature**          | **Before**                                                   | **After**                                                  |
| -------------------- | ------------------------------------------------------------ | ---------------------------------------------------------- |
| **Source**           | Generic `fetchLyricsFromWeb`                                 | Prioritized Netease Official API                           |
| **Language/Quality** | Results can be inconsistent (e.g., Cantonese or poor translations) | Consistently provides standard Chinese lyrics              |
| **Sync Precision**   | Supports 2-digit decimals only; 3-digit formats cause parsing errors | Full support for `[xx:xx.xxx]` 3-digit precision           |
| **Search Strategy**  | Single query                                                 | Dual-search strategy: "Title + Artist" followed by "Title" |

------

Before:
7aad3582be8a5a7e1b1f975c01fb7340
After:
762861a4e5855a632c72160d9114ba67
Lyric Source Upgrade: From Web Scraping to Netease Native API While the previous fetchLyricsFromWeb method often resulted in inaccurate matches or language inconsistencies (e.g., returning Cantonese instead of Mandarin), the new fetchLyricsFromNetease logic fetches lyrics directly from Netease's official database, ensuring higher quality and more standard results.

Before:
5e6fac0f11d254447c37d151f06eee61
After:
9e8688ac4a3dbf7817e6857fe5cfc852
Compatibility Fix: Supporting 3-digit Millisecond Timestamps [mm:ss.xxx] The original parser was limited to 2-digit centisecond formats. It now supports the [mm:ss.xxx] 3-digit millisecond format commonly used by Netease. By refactoring the regex and introducing a dynamic multiplier, the parser now ensures seamless and precise synchronization for all LRC formats.

*This description was translated and polished with the help of Gemini. I hope this contribution helps make the project even better!* 😊

### DIFF
--- a/boringNotch/managers/LyricsService.swift
+++ b/boringNotch/managers/LyricsService.swift
@@ -64,6 +64,23 @@ class LyricsService: ObservableObject {
                     return
                 }
             }
+
+            // Try NetEase lyrics first when playing from NetEase Music
+            if let bundleIdentifier = bundleIdentifier, bundleIdentifier.contains("com.netease.163music") {
+                guard !Task.isCancelled else { return }
+                let neteaseResult = await self.fetchLyricsFromNetease(title: title, artist: artist)
+
+                guard !Task.isCancelled else { return }
+                await MainActor.run {
+                    self.currentLyrics = neteaseResult.plain
+                    self.syncedLyrics = neteaseResult.synced
+                    self.isFetchingLyrics = false
+                    if !neteaseResult.plain.isEmpty {
+                        self.lyricsCache[cacheKey] = neteaseResult
+                    }
+                }
+                return
+            }
             
             // Fallback to web
             guard !Task.isCancelled else { return }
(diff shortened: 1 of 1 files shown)


LZWcc:
I've completed the cleanup of this Pull Request. Here is a summary of the updates:

- Target Branch: Corrected the base branch from main to dev.

- Commit Cleanup: Squashed the previous history into a single clean commit to maintain a tidy project history.

Alexander5015:
could you please resolve the merge conflicts?

LZWcc:
> could you please resolve the merge conflicts?

Sure, working on it

LZWcc:
> could you please resolve the merge conflicts?

In a previous commit I already addressed a compatibility issue with NetEase LRC timestamps.

The original parser only supported the `[mm:ss.xx]` centisecond format, while NetEase commonly uses `[mm:ss.xxx]` millisecond timestamps.  
That change extended the regex and introduced a dynamic multiplier so the parser could correctly support `[mm:ss.xxx]`.

While testing NetEase lyrics further, I discovered another edge case.

Some NetEase LRC lines use the following timestamp format:  [00:00.00-1]

Because the `-1` suffix was not part of the original regex, the timestamp was not fully consumed during parsing, leaving part of the tag in the rendered lyrics text
This commit slightly extends the timestamp regex to support the NetEase suffix format while keeping compatibility with existing formats.

After this change:

- `[00:00.00-1]` timestamps are parsed correctly
- no leftover timestamp fragments appear in the lyrics text
- existing `[mm:ss]`, `[mm:ss.xx]`, and `[mm:ss.xxx]` formats remain fully supported

Before:
89c8ce9a3482e5c77abc3546090b680b
After:
4e0963b38c5c092c9b7b0d778d19ef51

boringNotch/managers/LyricsService.swift
```diff
@@ -157,6 +174,72 @@ class LyricsService: ObservableObject {
...
+            request.setValue("https://music.163.com", forHTTPHeaderField: "Referer")
+
+            do {
+                let (searchData, _) = try await URLSession.shared.data(for: request)
+                if let json = try JSONSerialization.jsonObject(with: searchData) as? [String: Any],
+                   let result = json["result"] as? [String: Any],
+                   let songs = result["songs"] as? [[String: Any]],
+                   let firstSong = songs.first {
+                    let rawId = firstSong["id"]
+                    if let songId = (rawId as? Int) ?? (rawId as? NSNumber)?.intValue {
```
Alexander5015, line 208: Should this use the scoring done by the existing LRCLib path to ensure the best result is used rather than the first one, or does it always return the correct result first?

LZWcc: I tested this with several NetEase Music tracks, including title + artist searches and title-only fallback cases. In those tests, the NetEase search endpoint consistently returned the intended song as the first result, so I kept the NetEase path simple by using the first result directly.

I’m still not very experienced with Swift, but if you think matching multiple candidates is necessary here, I can try adapting the NetEase search results so they can reuse the existing `findBestMatch` logic.

PR #962, opened 2026-01-09
Fix play pause doubleclick in plexamp

benjaminfrombe:
in some apps (like Plexamp) play/pause button toggling was not working


benjaminfrombe:
wrong branch

PR #964, opened 2026-01-09
Use explicit play/pause for Now Playing - fix plexamp control 

benjaminfrombe:
Play/Pause was acting likt it was double-click very fast in apps like plexamp


Alexander5015:
Interesting, does it work with the keyboard play/pause key? I think that it uses the same toggle function.

benjaminfrombe:
nevermind, I noticed it's something in plexamp, using the macOS system controls has the same behaviour (sometimes it doesn't work, or act like it doubleclicks).

Alexander5015:
You should report this to them. Since the macOS system controls also don’t work, I’d assume they’d want to address it quickly.

PR #975, opened 2026-01-13
Feat/Add 'Clear All' button to shelf

Ben-Lloyd05:
1. Implements a button that allows users to remove all items stored in the item shelf.
- Also implements right-click functionality that does the same.

2. Implements a number placed above the shelf icon signifying the number of files currently placed in the shelf.

image


boringNotch/Localizable.xcstrings
```diff
@@ -5202,6 +5202,107 @@
         }
       }
     },
+    "Clear All" : {
```
Alexander5015: PRs should not include translations, this is managed separately

Alexander5015 (changes requested):
Please rebase your changes onto `dev`, and set the target branch to `dev` as well.

boringNotch/components/Tabs/TabSelectionView.swift
```diff
@@ -21,32 +21,48 @@ let tabs = [
...
+            // Badge for shelf file count
+            if !shelfVM.isEmpty {
+                Text("\(shelfVM.items.count)")
+                    .font(.system(size: 10, weight: .bold))
+                    .foregroundStyle(.white)
+                    .frame(width: 18, height: 18)
+                    .background(Color.accentColor)
+                    .clipShape(Circle())
+                    .offset(x: 5, y: -5)
             }
```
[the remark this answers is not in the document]
Ben-Lloyd05: > The badge is positioned globally over the entire HStack, which means it will appear at the top-trailing of both tabs. Since the badge should likely only appear on the Shelf tab, consider wrapping the ZStack alignment logic differently or conditionally showing the badge only when the Shelf tab is visible to avoid confusion about which tab it belongs to.

For reference, Copilot's suggestion isn't valid as this leads to clipping of the badge element.

boringNotch/Localizable.xcstrings
Ben-Lloyd05: Both have now been resolved

Alexander5015 (changes requested):
Please update to fix merge conflicts

Ben-Lloyd05:
> Please update to fix merge conflicts

Fixed

boringNotch/components/Tabs/TabSelectionView.swift
Alexander5015, line 1: Tab selection is intended to be generalized, don't hardcode shelf state and shelf UI into the TabSelectionView

boringNotch/components/Tabs/TabSelectionView.swift
```diff
@@ -21,32 +21,48 @@ let tabs = [
...
+                                    .hidden()
+                            }
                         }
                     }
+                }
+            }
+            .clipShape(Capsule())
+            
+            // Badge for shelf file count
+            if !shelfVM.isEmpty {
```
Alexander5015, line 56: Badge is also clipped some of the time. I know this has more to do with a layout issue (I think that shouldn't depend on closed notch height does). But until that is resolved, the badge won't show up properly for everyone and shouldn't be added. It also might be nice if this badge was optional.

Image

boringNotch/components/Shelf/Views/ShelfView.swift
```diff
@@ -93,18 +102,47 @@ struct ShelfView: View {
...
-                .onDrop(of: [.fileURL, .url, .utf8PlainText, .plainText, .data], isTargeted: $vm.dragDetectorTargeting) { providers in
-                    handleDrop(providers: providers)
+                    .padding(-spacing)
+                    .scrollIndicators(.never)
+                    .onDrop(of: [.fileURL, .url, .utf8PlainText, .plainText, .data], isTargeted: $vm.dragDetectorTargeting) { providers in
+                        handleDrop(providers: providers)
+                    }
+                    
+                    // Clear All button
+                    Button(action: {
```
Alexander5015, line 121: I'm not sure about this. given that it is always floating in the file view, we may want some kind of confirmation flow to prevent accidental deletion, which seems quite easy, and we may also want this floating button to be enabled/disabled by a setting in the shelf settings.

Alexander5015 (changes requested):

lasersushi (changes requested):
The user @princip requested an addition to this pr in issue #1248:
"PR https://github.com/TheBoredTeam/boring.notch/pull/975 looks like a great addition, especially the 'Clear All' button and the file counter. However, it doesn't look like that PR includes the keyboard shortcuts (Backspace) for removing specific items or selections. I’d still love to see those implemented as well to make the shelf management feel even more efficient for macOS."

It would also be nice if the shortcut could be customized. I am happy to write the code for this if you want.

Alexander5015:
Closing because this is stale and the feedback hasn't been addressed. If you are still interested in working on this, feel free to open another PR.

Issue #969, opened 2026-01-12
[FEATURE] Less notch area when inactive


hexart:
### Is your feature request related to a problem?

When Boring Notch is inactive, it displays the default notch size on external displays as well. I would like it to follow NotchNook’s behavior: show a smaller area when inactive to indicate the app’s presence, while occupying as little screen space as possible.

### Describe the solution you'd like

Our boring notch:
<img width="241" height="74" alt="Image" src="https://github.com/user-attachments/assets/c09e0cab-05d5-4c0c-9710-4d84c5f44722" />

NotchNook:
<img width="281" height="48" alt="Image" src="https://github.com/user-attachments/assets/24e371e4-bbde-423c-b5ba-497f7b91fb7b" />

### Willingness to contribute

- [ ] Yes, I can write the code for this feature
- [ ] Yes, I can help with design/mockups

### Pre-submission checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have provided sufficient detail for the team to understand the request

selfsta:
working on it

hexart:
> working on it

wish we could make the bouncing effect & shadow effect smooth like Alcove.

PR #983, opened 2026-01-20
Feat: Compact inactive notch on external displays 

selfsta:
Issue #969 suggested making the idle notch smaller when inactive on external displays that don't have a notch to indicate the app’s presence, while occupying as little screen space. 

To implement this feature, I’ve added a new case to the ContentView file. This case checks for two new flags. To achieve this, I’ve made some modifications to the sizing/matters.swift file, reusing the logic used to detect a non-notch display. I’ve reorganized this logic into a getNotch function that sets a new global hasNotch flag to indicate the presence or absence of a physical notch. The rest of the functionality remains unchanged by these changes. Additionally, I’ve introduced a new constant flag nonNotchCompact in Constants.swift, which is tied to a new toggle in the SettingsView file. This toggle allows users to enable and disable this feature at their convenience. These changes only affect the notch when it’s idle and when the display lacks a notch. The remaining functionality remains unaffected.

Functionality showcase
example

New Toggle in Settings
Screenshot 2026-01-20 at 12 34 36 PM

While there’s potential to make the notch even smaller and more compact, pushing it further resulted in the elements responsible for drawing the notch breaking and separating, which affected the overall presentation.


selfsta:
I realized after submitting this that PR #924 addresses the same feature. I missed that earlier PR before I started (this is my first contribution to the project and to open source in general), but I’ve since reviewed that implementation as well. 

My approach focuses on minimizing the footprint of the changes by reusing existing logic, which I believe may be easier to maintain long-term. I also noticed a potential issue in #924 where the toggle doesn't currently turn off the feature once enabled; I’ve ensured the logic here handles both states correctly.

I'll defer to your judgment on which implementation aligns better with the project’s standards. I’m happy to either continue with this PR or help port my fixes over to #924.

boringNotch/Localizable.xcstrings
```diff
@@ -23616,5 +23620,5 @@
       }
     }
   },
-  "version" : "1.0"
```
Alexander5015, line 23619: I don't think this change is correct. It should still be version 1.0, because i think this indicates the versioning of xcstrings format, which hasn't changed.

boringNotch/sizing/matters.swift
```diff
@@ -9,6 +9,8 @@ import Defaults
 import Foundation
 import SwiftUI
 
+var hasNotch: Bool = false // Added global hasNotch variable
```
Alexander5015, line 12: Shouldn't have a global var for this, multiple monitor setups will have both non-notch and notch displays

boringNotch/sizing/matters.swift
```diff
@@ -36,6 +38,18 @@ enum MusicPlayerImageSizes {
     return nil
 }
 
+// Moved this logic out of getClosedNotchSize
+@MainActor func getNotch(screen: NSScreen) {
```
Alexander5015, line 42: Instead of setting a global variable here, this should return a value scoped to the specific display, or possibly be implemented as an extension on `NSScreen` or attached to the UUID. I’m not fully sure on the best pattern yet, but this should be per-display rather than global. It’s worth researching a clean per-display approach, and potentially moving this out of `sizing`, since it could be a more general display-level utility that `sizing` calls into.

Also, I think this should have explicitly set the variable to `false` rather than leaving it unchanged, since it may have been set to `true` earlier.

boringNotch/Localizable.xcstrings
selfsta: I believe this was changed by mistake when I opened the project on Xcode or forking the repo

Alexander5015:
See #1008

selfsta:
Haven't been able to do further work on this, will take a look as soon as possible.

Alexander5015:
Basically I used your PR and made my own commits on top in #1008 and merged it in.

PR #1035, opened 2026-02-12
OSD Improvements

Alexander5015:
This pull request introduces new integrations with the Lunar app and BetterDisplay app to enable real-time brightness event streaming and control, along with a refactor to rename "HUD" (Heads-Up Display) references to "OSD" (On-Screen Display) throughout the project.

### DIFF
--- a/boringNotch/components/OSD/Managers/BetterDisplayManager.swift
+++ b/boringNotch/components/OSD/Managers/BetterDisplayManager.swift
@@ -12,6 +12,30 @@ import CoreGraphics
 final class BetterDisplayManager {
     static let shared = BetterDisplayManager()
     
+    let betterDisplayBundleIdentifier = "pro.betterdisplay.BetterDisplay"
+
+    private enum ControlTarget {
+        case brightness
+        case volume
+        case mute
+        case other(String)
+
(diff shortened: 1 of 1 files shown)


boringNotch/components/Settings/ViewModels/OSDSettingsViewModel.swift
mlbonniec, line 1: Why not using the `@Observable` macro?

boringNotch/components/OSD/Managers/BetterDisplayManager.swift
```diff
@@ -0,0 +1,263 @@
...
+            return
+        }
+        
+        // Determine type from controlTarget or systemIconID
+        let target = osd.controlTarget ?? ""
+        let iconID = osd.systemIconID ?? -1
+        
+        let isBrightness = target.contains("Brightness") || target.contains("brightness") || iconID == 1
+        let isVolume = target == "volume" || iconID == 3
+        let isMute = target == "mute" || iconID == 4
```
mlbonniec: Maybe you could use some constants here instead of raw values? Or directly an enum for `controlTarget`

Issue #1002, opened 2026-01-30
[FEATURE] Option for horizontal flip in mirror


dev-shreyansh:
### Is your feature request related to a problem?

No feature to flip the mirror

### Describe the solution you'd like

Flipping the feature would be great addition.

### Willingness to contribute

- [x] Yes, I can write the code for this feature
- [ ] Yes, I can help with design/mockups

### Pre-submission checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have provided sufficient detail for the team to understand the request

Alexander5015:
Sounds good, feel free to open a PR for this, we already need a section for other camera/mirror related settings anyway. If you want, you can also add in a setting to choose camera source here.

PR #1049, opened 2026-02-21
Added mirror toggle to boring mirror settings

anmolmalhotra97:
**Add Mirror Toggle to Mirror Settings**

Why was it changed:
- To provide users with more customization over their webcam feed, allowing them to choose between a mirrored or true-to-life perspective.

**Related Issues**
Label: Enhancement -> https://github.com/TheBoredTeam/boring.notch/issues/1002

**Screenshots / Screen Recordings**

https://github.com/user-attachments/assets/0003244e-e849-4d2a-b3f5-ea7824465762

Screen Recording
<br>
image
Notch Open
<br>
image
Webcam Mode
<br>
Screenshot 2026-02-24 at 01 25 15
Flip button not visible in the webcam feed, since disabled in webcam settings
<br>
Screenshot 2026-02-24 at 01 25 26
Flip button is visible in the webcam feed, since enabled in the webcam settings
<br>
image
Normal Webcam (Non-flipped/mirrored)
Note: Bottom right corner has a flip/mirror button
<br>
image
Flipped Webcam Feed
<br>
image
Regular Webcam Feed (click to Flip Button)
<br>
image
Mirrored Webcam Feed (click to return to normal)

### DIFF
--- a/boringNotch/components/Settings/SettingsView.swift
+++ b/boringNotch/components/Settings/SettingsView.swift
@@ -43,8 +43,8 @@ struct SettingsView: View {
                 NavigationLink(value: "Shelf") {
                     Label("Shelf", systemImage: "books.vertical")
                 }
-                NavigationLink(value: "Webcam") {
-                    Label("Webcam", systemImage: "camera")
+                NavigationLink(value: "Mirror") {
+                    Label("Mirror", systemImage: "camera")
                 }
                 NavigationLink(value: "Shortcuts") {
                     Label("Shortcuts", systemImage: "keyboard")
@@ -77,8 +77,8 @@ struct SettingsView: View {
                     Charge()
                 case "Shelf":
                     Shelf()
-                case "Webcam":
-                    WebcamSettings()
+                case "Mirror":
+                    MirrorSettings()
                 case "Shortcuts":
                     Shortcuts()
                 case "Advanced":
--- a/boringNotch/components/Settings/Views/WebcamSettingsView.swift
+++ b/boringNotch/components/Settings/Views/WebcamSettingsView.swift
@@ -1,26 +1,50 @@
 //
-//  WebcamSettings.swift
+//  WebcamSettingsView.swift
 //  boringNotch
 //
 //  Created by Anmol Malhotra on 2026-02-24.
 //
 
+import AVFoundation
 import SwiftUI
 import Defaults
 
-struct WebcamSettings: View {
-
-    @Default(.enableFlipWebcamToggle) private var enableFlipWebcamToggle
+struct MirrorSettings: View {
+    @Default(.showMirror) private var showMirror
+    @Default(.isMirrored) private var isMirrored
+    @Default(.mirrorShape) private var mirrorShape
 
     var body: some View {
         Form {
-            Defaults.Toggle(key: .enableFlipWebcamToggle) {
-                Text("Enable toggle to flip webcam")
+            Section {
+                Defaults.Toggle(key: .showMirror) {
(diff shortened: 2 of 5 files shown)


anmolmalhotra97:
Hi @Alexander5015 ,
I have added a mirror toggle feature to the webcam feed.

Please review this PR and let me know if you have any recommendations.

boringNotch/components/Webcam/WebcamView.swift
```diff
@@ -44,6 +46,22 @@ struct CameraPreviewView: View {
                     }
                 }
             }
+            // The mirror toggle button should only be visible if the webcam session is running
```
Alexander5015: I think we can make this a setting in Settings, unless you believe there’s a strong case for keeping it here.

boringNotch/components/Webcam/WebcamView.swift
```diff
@@ -15,13 +15,15 @@ struct CameraPreviewView: View {
     
     // Track if authorization request is in progress to avoid multiple requests
     @State private var isRequestingAuthorization: Bool = false
+    // Track the current state of mirror effect and the mirror icon in camera preview
+    @State private var isMirrored: Bool = true
```
Alexander5015: This should be stored in Defaults to make it persistent

Alexander5015 (changes requested):
I’ve left some feedback. Also, as a side note: as additional changes may land in dev before this is merged, please periodically rebase the branch on dev to keep it updated.

boringNotch/components/Webcam/WebcamView.swift
anmolmalhotra97: Hey @Alexander5015 ,

Would something like this work?

Image

boringNotch/components/Webcam/WebcamView.swift
anmolmalhotra97: I have updated the code to cater to this suggestion.

anmolmalhotra97:
@Alexander5015, I have updated the codebase as per your suggestions. 
I have also updated the PR description.

Please review and provide feedback.

boringNotch/components/Settings/Views/WebcamSettingsView.swift
Alexander5015, line 1: Move all mirror settings here. I think its the setting to enable the mirror and the mirror shape.

boringNotch/components/Settings/SettingsView.swift
```diff
@@ -43,6 +43,9 @@ struct SettingsView: View {
                 NavigationLink(value: "Shelf") {
                     Label("Shelf", systemImage: "books.vertical")
                 }
+                NavigationLink(value: "Webcam") {
```
Alexander5015: Rename to Mirror to match the feature name

boringNotch/components/Webcam/WebcamView.swift
```diff
@@ -15,16 +15,36 @@ struct CameraPreviewView: View {
     
     // Track if authorization request is in progress to avoid multiple requests
     @State private var isRequestingAuthorization: Bool = false
-
+    // Track the current state of mirror effect and the mirror icon in camera preview
+    @Default(.isMirrored) private var isMirrored
+    @Default(.enableFlipWebcamToggle) private var enableFlipWebcamToggle
```
Alexander5015: I am open to discussion about this, but I don't think we need a toggle button overlay at all. My suggestion was that rather than having a visible toggle button, we just make this a setting in settings as I don't think this will be frequently changed.

PR #1050, opened 2026-02-22
feat: add lyrics column layout with scroll browsing and calendar inli…

rajuptvs:
- Add 3-line synced lyrics column (prev/current/next) beside compact music player when LRC lyrics are available; toggled via Settings or tapping the inline lyric line
- Lyrics scroll browsing: mouse scroll gesture moves through lines with a 2-second auto-snap back; uses NSEvent local monitor to reliably capture scroll events behind SwiftUI content layers
- Calendar moves into the right panel (replacing lyrics column) toggled from a new capsule button in BoringHeader; separator line collapses the panel
- Lyrics disk cache at ~/Library/Application Support/boringNotch/lyrics_cache.json to avoid redundant LRCLIB fetches; auto-retry once on network failure
- Retry button shown in lyrics column when lyrics are genuinely unavailable
- Add lyricsColumnLayout key to Constants and sub-toggle in Settings

https://github.com/user-attachments/assets/18d6891e-7127-47c5-884a-5a627dc729a3

### DIFF
--- a/boringNotch/boringNotch.entitlements
+++ b/boringNotch/boringNotch.entitlements
@@ -4,6 +4,8 @@
 <dict>
 	<key>com.apple.security.app-sandbox</key>
 	<true/>
+	<key>com.apple.security.cs.disable-library-validation</key>
+	<true/>
 	<key>com.apple.security.automation.apple-events</key>
 	<true/>
 	<key>com.apple.security.device.camera</key>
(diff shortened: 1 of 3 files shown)


theboringhumane:
Looks good overall. Instead of enabling lyrics by default on closed notch, could you add a top toggle button (similar to the calendar one) to show/hide lyrics, and reuse the calendar icon for this?

rajuptvs:
> Looks good overall. Instead of enabling lyrics by default on closed notch, could you add a top toggle button (similar to the calendar one) to show/hide lyrics, and reuse the calendar icon for this?

I should be able to add that but don't know if that would be a really good ux decision. Adding an extra click, might prevent users in general to not use the feature entirely (on the flip side can reduce unneccesary calls to fetch lyrics)

Alexander5015:
please base your changes of off dev and fix conflicts

PR #1053, opened 2026-02-25
feat: make drag-and-drop trigger area configurable

tornado-bunk:
## Summary
This PR replaces the previous binary toggle for expanded drag detection with a more flexible 3-option configuration for the shelf drag-and-drop trigger area.

## Changes
- **New Configuration**: Added `DragDetectionArea` enum with three states:
    - `Disabled`: Completely removes global drag monitoring.
    - `Expanded area`: Uses the full open notch dimensions (original behavior).
    - `Notch only`: Restricts detection to the actual notch area.
- **UI Update**: Replaced the "Expanded drag detection" toggle with a Picker in the Shelf settings.

## Test plan
1. Open Settings -> Shelf.
2. Change "Drag detection area" to "Notch only".
3. Verify that dragging a file onto the closed notch (even when expanded by music) triggers the shelf.
4. Verify that dragging a file near the corners (outside the notch area) does not trigger it.
5. Change to "Disabled" and verify no shelf opens on drag.
Screenshot 2026-02-25 alle 01 41 26

### DIFF
--- a/boringNotch/boringNotchApp.swift
+++ b/boringNotch/boringNotchApp.swift
@@ -310,17 +310,20 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     func applicationDidFinishLaunching(_ notification: Notification) {
         
         // One-time migration from legacy `expandedDragDetection` to `dragDetectionArea`
-        if !Defaults.hasKey(.dragDetectionArea) {
-            let expandedDragDetectionEnabled = Defaults[.expandedDragDetection]
-            
-            if expandedDragDetectionEnabled == false {
-                Defaults[.dragDetectionArea] = .disabled
-            } else {
-                Defaults[.dragDetectionArea] = .openNotch
+        // Check if the new key is NOT set yet (using UserDefaults directly to check for existence)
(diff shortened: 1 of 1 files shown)


Alexander5015:
Is there any use case for having the drag monitoring disabled but the shelf enabled, or should it just disable drag monitoring when the shelf is disabled?

tornado-bunk:
Hi! I'm closing this PR for now to refine the implementation. After some testing and your feedback, I realized that the 'Disabled' option is indeed redundant since we already have a general 'Enable Shelf' toggle. I'll open a new, cleaner PR shortly that focuses only on the choice between 'Expanded area' and 'Notch only', without the extra complexity. Sorry for the noise!

Alexander5015:
No worries about th noise, I appreciate you taking time to work on this!

I think that the drag monitoring may not get disabled properly when shelf is disabled? Idk, I think there was some issue that was reported somewhere about the shelf not being easily disabled, so that might be something you can incude as well.

tornado-bunk:
Thanks for pointing that out! I'll definitely look into the drag monitoring behavior to see if it’s actually staying active when the shelf is disabled. If I find anything, I'll make sure to include a fix in the next PR.

PR #1056, opened 2026-02-25
feat: make drag-and-drop trigger area configurable & fix shelf disable behavior

tornado-bunk:
This PR addresses some issues regarding the Shelf's drag-and-drop functionality. 

First, it replaces the binary "Expanded drag detection" toggle with a more flexible `DragDetectionArea` configuration, allowing users to choose between the expanded area (full open notch size) or restricting detection strictly to the notch only.

Second, it fixes a bug where the Shelf would continue to detect drag events even when the "Enable shelf" setting was disabled. Previously, the global drag detectors and the `onDrop` modified in `ContentView` did not respect the disabled state of the shelf.

### Changes
- **New Feature**: Added `DragDetectionArea` enum and picker in Settings -> Shelf. Options: "Expanded area" (default) and "Notch only".
- **Bug Fix**: Updated `setupDragDetectors()` in `boringNotchApp.swift` to check `Defaults[.boringShelf]` before starting monitors.
- **Bug Fix**: Modified `ContentView.swift`'s `dragDetector` and general `onDrop` delegate to be conditional based on both `boringShelf` and `expandedDragDetection` settings.
- **Refactor**: Added `boringShelfChanged` notification to reactively update drag detector state without app restart.

### Test Plan
1. **Configuration Test**:
   - Go to Settings -> Shelf.
   - Set "Detection area" to "Notch only".
   - Drag a file to the physical notch -> Shelf should open.
   - Drag a file near the notch corners (outside physical notch) -> Shelf should NOT open.

2. **Disable Shelf Test**:
   - Toggle "Enable shelf" OFF.
   - Drag a file to the notch -> Shelf should NOT open.

3. **Disable Drag Detection Test**:
   - Toggle "Enable shelf" ON, "Enable drag detection" OFF.
   - Drag a file to the notch -> Shelf should NOT open.

### Screenshots
Screenshot 2026-02-25 alle 20 43 22

### Screen Recordings
https://github.com/user-attachments/assets/a07416f5-d2c6-4b67-b644-da93e76a293a


boringNotch/observers/DragDetector.swift
```diff
@@ -46,7 +46,7 @@ final class DragDetector {
             .string
         ]
         let isValid = dragPasteboard.pasteboardItems?.allSatisfy { item in
-            item.types.allSatisfy { validTypes.contains($0) }
+            item.types.contains { validTypes.contains($0) }
```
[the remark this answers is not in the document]
tornado-bunk: Yes, changing from allSatisfy to contains is intentional.
Pasteboard items (e.g. from Finder) usually include many extra types (dyn.*, com.apple.finder.*, etc.) in addition to .fileURL / URL / string.
With allSatisfy, a single non-whitelisted type caused us to reject otherwise valid drags, which is why drag detection stopped working reliably after the first drag during my tests.

boringNotch/components/Settings/Views/ShelfSettingsView.swift
```diff
@@ -25,18 +27,36 @@ struct Shelf: View {
...
+                        Text(area.rawValue).tag(area)
+                    }
+                }
+                .disabled(!expandedDragDetection)
+                .onChange(of: dragDetectionArea) {
+                    NotificationCenter.default.post(
+                        name: Notification.Name.expandedDragDetectionChanged,
+                        object: nil
+                    )
+                }
```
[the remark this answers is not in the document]
tornado-bunk: When expandedDragDetection is false we early‑return from setupDragDetectors(), so dragDetectionArea isn’t used at runtime; the disabled picker only shows (and preserves) which area will be used once detection is turned back on.

boringNotch/ContentView.swift
```diff
@@ -394,7 +397,22 @@ struct ContentView: View {
...
+            set: { newValue in
+                // Only update the actual model IF shelf AND detection are enabled
+                if boringShelf && expandedDragDetection {
+                    vm.generalDropTargeting = newValue
+                } else {
+                    // If disabled, force target to false to be safe
+                    vm.generalDropTargeting = false
+                    }
+                }
+            )))
```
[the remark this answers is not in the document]
tornado-bunk: fixed

boringNotch/Localizable.xcstrings
```diff
@@ -20936,5 +20945,5 @@
       }
     }
   },
-  "version" : "1.0"
+  "version" : "1.1"
```
Alexander5015: This should still be version 1.0 I believe. I'm guessing this is an AI edit, generally you shouldn't allow or accept AI edits to the `Localizable.xcstring` file, it is an autogenerated file based on the strings used in the application, and will automatically add the necessary keys when the code is updated after the next build.
```suggestion
  "version" : "1.0"
```

boringNotch/boringNotchApp.swift
```diff
@@ -205,19 +205,29 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
     private func setupDragDetectorForScreen(_ screen: NSScreen) {
```
Alexander5015, line 206: Doesn't the existing drop target already cover the closed notch? In this case a global monitor shouldn't be necessary. In general i would like to avoid the global monitor, since it false positives with Safari tabs and is probably less performant than a drop target.

boringNotch/Localizable.xcstrings
tornado-bunk: Thanks for the fix..
This is my first time with Swift and I used some AI assistance to help with the implementation. I just didn't realize that specific metadata file was managed by Xcode. Thanks!

boringNotch/components/Settings/Views/ShelfSettingsView.swift
```diff
@@ -25,18 +27,36 @@ struct Shelf: View {
...
+                        name: Notification.Name.boringShelfChanged,
+                        object: nil
+                    )
+                }
                 Defaults.Toggle(key: .openShelfByDefault) {
                     Text("Open shelf by default if items are present")
                 }
                 Defaults.Toggle(key: .expandedDragDetection) {
-                    Text("Expanded drag detection area")
+                    Text("Enable drag detection")
```
Alexander5015, line 40: Creating a new setting while still using an older variable is generally bad practice. In the current version, this value only disabled drag detection outside of the closed notch window, but now it is being used to disable all drag detection. If this code gets deployed, it will apply the old setting value to the new unrelated setting, causing unexpected behaviour for users by disabling dragging entirely.

boringNotch/components/Settings/Views/ShelfSettingsView.swift
```diff
@@ -25,18 +27,36 @@ struct Shelf: View {
...
                     )
                 }
+                Picker("Detection area", selection: $dragDetectionArea) {
+                    ForEach(DragDetectionArea.allCases) { area in
+                        Text(area.rawValue).tag(area)
+                    }
+                }
+                .disabled(!expandedDragDetection)
+                .onChange(of: dragDetectionArea) {
+                    NotificationCenter.default.post(
```
Alexander5015, line 55: While we eventually need to move away from using notifications for these settings, I’m not requesting a change here since this pattern is still consistent.

boringNotch/Localizable.xcstrings
Alexander5015: Also btw, you need to click "Commit suggestion" if you want to actually apply the change. I unresolved the suggestion and commited it.

boringNotch/components/Settings/Views/ShelfSettingsView.swift
theboringhumane: for now this is good.

PR #1057, opened 2026-02-26
Dapper Crab 🎩🦀

Alexander5015:


z76k:
when will it be merged

BoringNotchXPCHelper/MessagesSender.swift
```diff
@@ -0,0 +1,154 @@
...
+                        try
+                            if (name of p as string) is equal to targetName then
+                                set matched to true
+                                exit repeat
+                            end if
+                        end try
+                    end repeat
+                    if matched then
+                        if (count of participants of c) is 1 then
+                            send replyText to c
```
theboringhumane: **[P1] Resolve the originating Messages conversation before sending**

After a notification banner expires, `SystemNotificationManager.performReply` passes the notification's display name into this fallback. With two one-to-one conversations whose participants share that name, this loop sends immediately to the first match and returns without checking the other conversations. The group fallback also picks the first matching group. A private reply can therefore go to the wrong person or thread. The later bare-participant ambiguity check cannot protect these earlier sends. Require a verified conversation identifier; when it cannot be established, preserve the draft and hand off instead of guessing.

Verified against commit `f4372f511335e91cfbb5b77f4ea5822c5691c55d` by tracing the expired-banner fallback and AppleScript branches; no messages were sent during review.

boringNotch/helpers/OTPDetector.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    private static func charBefore(_ range: NSRange, in ns: NSString) -> Character? {
+        guard range.location > 0 else { return nil }
+        return Character(UnicodeScalar(ns.character(at: range.location - 1))!)
+    }
+
+    private static func charAfter(_ range: NSRange, in ns: NSString) -> Character? {
+        let end = range.location + range.length
+        guard end < ns.length else { return nil }
+        return Character(UnicodeScalar(ns.character(at: end))!)
```
theboringhumane: **[P1] Read a complete character at OTP boundaries**

A notification containing `Code 1234🔒` reaches this force unwrap while rendering its detected code. `NSString.character(at:)` returns the emoji's high UTF-16 surrogate (`0xD83D`), which is not a valid standalone Unicode scalar, so `UnicodeScalar(...)` is nil and the app crashes. An emoji before the digits similarly fails in `charBefore` at line 113. Decode the complete adjacent character, or handle surrogate units without force-unwrapping.

An isolated Swift/Foundation check confirmed that both adjacent surrogate conversions return nil; ordinary `Your code is 123456` remains a valid control case.

boringNotch/models/Constants.swift
```diff
@@ -36,47 +29,206 @@ enum HideNotchOption: String, Defaults.Serializable {
...
 }
 
 // Media controller types for selection in settings
 enum MediaControllerType: String, CaseIterable, Identifiable, Defaults.Serializable {
-    case nowPlaying = "Now Playing"
-    case appleMusic = "Apple Music"
-    case spotify = "Spotify"
-    case youtubeMusic = "YouTube Music"
+    case nowPlaying
+    case appleMusic
```
theboringhumane, line 126: **[P2] Migrate persisted values and keys before renaming them**

The previous release serialized the media source as values such as `Spotify` and `Apple Music`; these cases now serialize as `spotify` and `appleMusic`, while the `mediaController` preference key is unchanged. Existing selections no longer decode and fall back to Now Playing. The migration in `MusicManager` only records choice flags, so it does not preserve the chosen source. The same issue affects `SneakPeekStyle`, `OptionKeyAction`, and `SliderColorEnum`.

The HUD-to-OSD key renames in this file also need migration: an existing `hudReplacement = true` becomes an absent `osdReplacement`, whose default is false. Preserve the old serialized names or explicitly migrate the stored values and keys before their first read.

boringNotch/managers/MusicManager.swift
```diff
@@ -70,119 +125,359 @@ class MusicManager: ObservableObject {
...
+        effectiveMediaController = type
+        canFavoriteTrack = controller.supportsFavorite
+        volumeControlSupported = controller.supportsVolumeControl
+
+        controller.playbackStatePublisher
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self, controller] state in
+                guard let self,
+                      self.activeController === controller,
+                      state.lastUpdated != .distantPast
```
theboringhumane, line 374: **[P2] Accept legitimate playback reset states**

Quitting YouTube Music calls `resetPlaybackState()`, which publishes `PlaybackState(bundleIdentifier: ..., isPlaying: false)`. Its default `lastUpdated` is `.distantPast`, so this new filter drops the reset and leaves the terminated player's previous title, progress, and playing state visible. Inactive `updatePlaybackInfo()` emits the same ignored reset. Distinguish an uninitialized snapshot from a real reset, or stamp terminal reset events with a valid update time; apply the same rule to the other timestamp guard in this manager.

A focused Combine reproduction using the actual `PlaybackState` showed `isPlaying=true` after termination with this guard, versus `false` with the base behavior.

theboringhumane (commented):
Review of cached base `99900bf630a3d3e97fae079df2175993318d51f7` through commit `f4372f511335e91cfbb5b77f4ea5822c5691c55d`: 2 P1 and 8 P2 findings. Comments refer specifically to that reviewed commit. Focused Swift reproductions and source/caller checks were performed; a full Xcode build and live UI tests were not run. Current remote-head changes could not be revalidated from this environment.

boringNotch/components/OSD/Managers/XPC/BrightnessManager.swift
```diff
@@ -0,0 +1,184 @@
...
+		pendingDelta += delta
+		guard flushTask == nil else { return }
+		flushTask = Task { @MainActor in
+			defer { flushTask = nil }
+			while pendingDelta != 0 {
+				let delta = pendingDelta
+				pendingDelta = 0
+				// Compute from the cached value — the read-back RPC the old
+				// code did before every set doubles the cost per key press.
+				let target = max(0, min(1, rawBrightness + delta))
```
theboringhumane: **[P2] Keep keyboard brightness synchronized before applying deltas**

This cache is refreshed at initialization/failure or after this manager writes, so changing keyboard brightness through Control Center or automatic adjustment leaves `rawBrightness` stale. The next keypress can jump to the old level instead of adjusting the current hardware value. Also, `publish` schedules its update asynchronously: when another delta is already pending, the next loop iteration calculates from the preceding cached value and loses a repeat.

An isolated harness using the extracted manager and a stub XPC client produced hardware targets `[0.5625, 0.5625]` for two increases from `0.5`; the expected targets are `[0.5625, 0.625]`. After an external change to `0.875`, one decrease wrote `0.5` instead of `0.8125`. The base implementation passed these cases. Synchronize the working value and account for external changes before calculating relative adjustments.

boringNotch/observers/DragDetector.swift
```diff
@@ -45,7 +45,10 @@ final class DragDetector {
             NSPasteboard.PasteboardType(UTType.url.identifier),
             .string
         ]
-        return dragPasteboard.types?.contains(where: validTypes.contains) ?? false
+        let isValid = dragPasteboard.pasteboardItems?.allSatisfy { item in
+            item.types.allSatisfy { validTypes.contains($0) }
```
theboringhumane: **[P2] Accept a supported representation even when auxiliary types exist**

A valid formatted-text drag commonly advertises both `.string` and `.rtf`. Requiring every representation to appear in this three-type whitelist rejects that drag despite its usable string representation. `isContentDragging` stays false, so entering the expanded detection region never opens the shelf. Check whether each item offers a supported representation instead of rejecting its other advertised types.

An isolated check with standalone pasteboard items accepted plain text in both versions, but plain text plus RTF changed from accepted in the base to rejected here; the user's pasteboard was not touched.

boringNotch/components/Shelf/ViewModels/ShelfStateViewModel.swift
```diff
@@ -51,28 +60,7 @@ final class ShelfStateViewModel: ObservableObject {
...
-            
-            for (itemID, bookmarkData) in self.pendingBookmarkUpdates {
-                if let idx = self.items.firstIndex(where: { $0.id == itemID }),
-                   case .file = self.items[idx].kind {
-                    self.items[idx].kind = .file(bookmark: bookmarkData)
-                }
-            }
-            
-            self.pendingBookmarkUpdates.removeAll()
+            items[idx] = ShelfItem(kind: .file(bookmark: bookmark), isTemporary:  items[idx].isTemporary)
```
theboringhumane: **[P2] Preserve the item UUID when replacing a bookmark**

`ShelfItem` defaults its `id` parameter to a fresh UUID, so this replacement changes the item's identity. Renaming a file through the shelf calls this method and leaves selection pointing at the old ID. Refreshing a stale bookmark while dragging also invalidates the original `draggedItems` snapshot: auto-removal later matches the old ID and leaves the item behind. Pass the existing item ID into the replacement constructor.

An isolated harness using the extracted constructor/update/remove code confirmed `bookmark update preserves identity: false`, zero selected items after the update, and one item remaining after removal via the original drag snapshot.

boringNotch/managers/SystemNotificationManager.swift
```diff
@@ -0,0 +1,629 @@
...
+
+        // WhatsApp exposes no scripting interface, but its URL scheme can
+        // open a specific conversation with text pre-filled — far better
+        // than the clipboard, which leaves the user to find the chat and
+        // paste. Needs a phone number, which only comes from Contacts and
+        // only when it resolves unambiguously (see phoneNumber(for:)).
+        if notification.bundleID == "net.whatsapp.WhatsApp",
+           let sender = notification.sender,
+           let phone = await ContactAvatarManager.shared.phoneNumber(forContactNamed: sender),
+           let encoded = text.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed),
```
theboringhumane: **[P2] Encode the WhatsApp draft as a query value**

When an expired WhatsApp banner falls back to a resolved contact, a draft such as `Tom & Jerry` is encoded as `Tom%20&%20Jerry` because `.urlQueryAllowed` permits `&`. The resulting URL's `text` parameter is only `Tom `; the rest becomes another parameter. This path then dismisses the notification and clears the saved draft, losing the original text. Construct the URL using properly encoded query items so arbitrary draft content remains one value.

Swift/Foundation parsed the generated URL as `whatsapp://send?phone=1234567890&text=Tom%20&%20Jerry` and decoded its `text` value as `Tom `.

BoringNotchXPCHelper/NotificationWatcher.swift
```diff
@@ -0,0 +1,765 @@
...
+    /// keyboard focus even with the window parked off-screen (macOS does
+    /// not unfocus off-screen windows) — so the toggle, and the off-screen
+    /// park in hold(token:) that hides the expanded banner, both run only
+    /// while the notch is open. Defaults to false: never steal focus
+    /// before being told.
+    var notchOpen: Bool = false {
+        didSet {
+            if notchOpen, !oldValue {
+                // Re-arm the skip log for the next closed episode.
+                skipLogged.removeAll()
```
theboringhumane, line 73: **[P2] Park already-held banners on the closed-to-open transition**

A notification arriving while the notch is closed is held once in `SystemNotificationManager.add`, but `hold(token:)` returns before parking because `notchOpen` is false. Opening the notch only sends `notchOpened`; this transition clears the log set but never parks those existing holds. There is no subsequent app-side `holdNotification` call, despite the comment in `hold(token:)`. The next keep-alive performs the details action on the still-visible system banner, exposing its competing reply field while the notch composer is open. Park eligible existing holds on this transition before refreshing them.

Verified by tracing the arrival, notch-state, hold, and keep-alive call paths at the reviewed commit; live notification UI was not exercised.

boringNotch/managers/NotchWindowManager.swift
```diff
@@ -0,0 +1,390 @@
...
+                self?.handleDragEntersNotchRegion(onScreen: screen)
+            }
+        }
+
+        // In single-screen mode there may be no context entry yet — only
+        // multi-display mode registers per-screen contexts.
+        if contexts[uuid] != nil {
+            contexts[uuid]?.dragDetector = detector
+        } else {
+            contexts[uuid] = ScreenContext(viewModel: primaryViewModel, window: nil, dragDetector: detector)
```
theboringhumane: **[P2] Give each display its own view model when enabling all displays**

With expanded drag detection enabled, select display A in single-display mode, then select B, then enable Show on all displays. Each drag setup stores the same `primaryViewModel` under its screen UUID. Drag cleanup retains the contexts, and single-window cleanup also leaves them in place. Multi-display setup consequently builds both windows using that same model, whose `screenUUID` is B. Opening one notch opens both, and display A uses B's screen-specific state. Clear these single-display-only contexts or replace them with distinct per-screen models before multi-display window creation.

The source flow and an isolated Swift reference-semantics reproduction confirmed shared identity, A reporting screen UUID B, and opening A changing B. No real display settings were changed during review.

PR #1075, opened 2026-03-01
feat: introduce lunar calendar support

liWanr:
## 🔧Settings UI

- **Added "Calendar subtitle display" option** with three modes:
    - **Always default**: Keeps the original year display (e.g., "2026").
    - **Always alternate**: Permanently displays the alternate calendar.
    - **Tap to switch**: Allows users to toggle between the year and the alternate calendar by tapping the subtitle.
- **Added conditional "Alternate Calendar" Picker**: This picker appears only when "Always alternate" or "Tap to switch" is selected.

    > **Note on Localization**: Currently, **this feature only supports the Lunar Calendar**. Since there is no standard or intuitive way to express Lunar dates in non-Chinese languages (e.g., it defaults to strings with ambiguous meanings like `mo1 14` in English), **therefore the Lunar display is locked to Simplified Chinese** to ensure clarity and accuracy.

## 📅Calendar View

- **Layout Optimization**: Implemented a fixed-width hidden view strategy using `ZStack`. This ensures the UI remains rock-solid and prevents any layout "jitter" or shifting when toggling between the year and Lunar text.
- **State Persistence**: The toggle state (`calendarIsShowingAlternate`) is now persisted. Even after closing the window or restarting the app, it will remember the last selection in "Tap to switch" mode.

## 📃Localization

- Added localization objects for all new settings in `Localizable.xcstrings`.
- **Note**: I have not added translations for non-default languages (**including Simplified Chinese, which I am most familiar with**) yet, as **I believe** all translation work should be officially handled via **Crowdin** to maintain project standards.

## 📷Screenshots

<div align="center">

image

image

image

image

image

</div>

### DIFF
--- a/boringNotch/components/Calendar/BoringCalendar.swift
+++ b/boringNotch/components/Calendar/BoringCalendar.swift
@@ -233,6 +233,34 @@ struct CalendarView: View {
     @EnvironmentObject var vm: BoringViewModel
     @ObservedObject private var calendarManager = CalendarManager.shared
     @State private var selectedDate = Date()
+    @Default(.calendarSubtitleDisplayMode) var displayMode
+    @Default(.alternativeCalendar) var alternativeCalendar
+    @Default(.calendarIsShowingAlternate) var isShowingAlternate
+
+    private var shouldShowAlternate: Bool {
+        switch displayMode {
+        case .alwaysDefault:
+            return false
+        case .alwaysAlternate:
+            return true
+        case .tapToSwitch:
+            return isShowingAlternate
+        }
+    }
+
+    // A hidden view strictly for sizing
+    private func subtitleSizingView(date: Date) -> some View {
+        ZStack {
+            Text(date.formatted(.dateTime.year()))
+            switch alternativeCalendar {
+            case .lunar:
+                Text(date.formatted(.dateTime.lunar()))
+            }
+        }
+        .font(.title3)
+        .fontWeight(.light)
+        .fixedSize() // Ensure it takes up its natural calculated size
+    }
 
     var body: some View {
         VStack(spacing: 0) {
@@ -242,10 +270,34 @@ struct CalendarView: View {
                         .font(.title3)
                         .fontWeight(.semibold)
                         .foregroundColor(.white)
-                    Text(selectedDate.formatted(.dateTime.year()))
+                    
+                    ZStack(alignment: .leading) {
+                        // Invisible layer to reserve space for the widest possible content
+                        subtitleSizingView(date: selectedDate)
+                            .hidden()
+                            
+                        // Visible content
+                        Group {
+                            if shouldShowAlternate {
+                                switch alternativeCalendar {
+                                case .lunar:
+                                    Text(selectedDate.formatted(.dateTime.lunar()))
+                                }
+                            } else {
+                                Text(selectedDate.formatted(.dateTime.year()))
+                            }
+                        }
                         .font(.title3)
                         .fontWeight(.light)
                         .foregroundColor(Color(white: 0.65))
+                        .onTapGesture {
+                            if displayMode == .tapToSwitch {
+                                withAnimation(.smooth(duration: 0.2)) {
+                                    isShowingAlternate.toggle()
+                                }
+                            }
+                        }
+                    }
                 }
 
                 ZStack(alignment: .top) {
--- a/boringNotch/extensions/DataTypes+Extensions.swift
+++ b/boringNotch/extensions/DataTypes+Extensions.swift
@@ -42,6 +42,31 @@ extension Date {
     }
 }
 
+struct LunarDateStyle: FormatStyle {
+    func format(_ value: Date) -> String {
+        let formatter = DateFormatter()
+        formatter.calendar = Foundation.Calendar(identifier: .chinese)
+        formatter.locale = Locale(identifier: "zh_CN")
+        formatter.dateStyle = .long
+        formatter.timeStyle = .none
+        
+        let fullDate = formatter.string(from: value)
+        
+        if let yearRange = fullDate.range(of: "年") {
+            let extracted = String(fullDate[yearRange.upperBound...])
+            return extracted
+        }
+        
+        return fullDate
+    }
+}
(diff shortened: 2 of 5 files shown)


boringNotch/components/Calendar/BoringCalendar.swift
```diff
@@ -242,10 +242,17 @@ struct CalendarView: View {
                         .font(.title3)
                         .fontWeight(.semibold)
                         .foregroundColor(.white)
-                    Text(selectedDate.formatted(.dateTime.year()))
-                        .font(.title3)
-                        .fontWeight(.light)
-                        .foregroundColor(Color(white: 0.65))
+                    if Locale.current.identifier.hasPrefix("zh") {
```
Alexander5015: Make this a separate setting rather than tying the calendar directly to the locale. These should be decoupled, as users may want to choose their calendar independently of their language, especially since both the Gregorian and Chinese calendars are commonly used in China.

Ideally, we could support a primary calendar along with an optional alternate calendar, similar to Apple Calendar, but idk how to present both clearly within the limited available space.

Alexander5015 (changes requested):

boringNotch/components/Calendar/BoringCalendar.swift
liWanr: You’re absolutely right. To address the space constraint, I have a simple idea: we could add a setting that lets users choose between a fixed display or a clickable switch. Basically, when the user taps the date (e.g., "2026"), it swaps the content to the Lunar date, and vice versa.

I’m not sure if this approach is feasible, what do you think?

Alexander5015: I like it, looking forward to seeing what you come up with! You could have a picker with the options something like: [Always Default] [Always Alternate] [Tap to switch]

liWanr: Should I add the localization object myself, or will you be handling that?

Alexander5015: For the settings? you should do it. Any Text views you add will automatically manage the localization. For the picker, you will need to manage it similarly to the existing ones

boringNotch/models/Constants.swift
```diff
@@ -134,6 +134,38 @@ enum OSDControlSource: String, CaseIterable, Identifiable, Defaults.Serializable
...
+        case .alwaysAlternate:
+            return NSLocalizedString("Always alternate", comment: "")
+        case .tapToSwitch:
+            return NSLocalizedString("Tap to switch", comment: "")
+        }
+    }
+}
+
+enum AlternativeCalendarType: String, CaseIterable, Identifiable, Defaults.Serializable {
+    case lunar = "Lunar"
```
Alexander5015, line 157: Change 'lunar' to 'Chinese.' The Chinese calendar is actually lunisolar, and 'lunar' is too ambiguous since several other calendars use the lunar or lunisolar system.

liWanr:
# Change Description

## Settings UI

- **Added "Calendar subtitle display" option** with three modes:
    - **Always default**: Keeps the original year display (e.g., "2026").
    - **Always alternate**: Permanently displays the alternate calendar.
    - **Tap to switch**: Allows users to toggle between the year and the alternate calendar by tapping the subtitle.
- **Added conditional "Alternate Calendar" Picker**: This picker appears only when "Always alternate" or "Tap to switch" is selected.

    > **Note on Localization**: Currently, **this feature only supports the Lunar Calendar**. Since there is no standard or intuitive way to express Lunar dates in non-Chinese languages (e.g., it defaults to strings with ambiguous meanings like `mo1 14` in English), **therefore the Lunar display is locked to Simplified Chinese** to ensure clarity and accuracy.

## Calendar View

- **Layout Optimization**: Implemented a fixed-width hidden view strategy using `ZStack`. This ensures the UI remains rock-solid and prevents any layout "jitter" or shifting when toggling between the year and Lunar text.
- **State Persistence**: The toggle state (`calendarIsShowingAlternate`) is now persisted. Even after closing the window or restarting the app, it will remember the last selection in "Tap to switch" mode.

## Localization

- Added localization objects for all new settings in `Localizable.xcstrings`.
- **Note**: I have not added translations for non-default languages (**including Simplified Chinese, which I am most familiar with**) yet, as **I believe** all translation work should be officially handled via **Crowdin** to maintain project standards.

## Screenshots

<div align="center">

image

image

image

image

image

</div>

liWanr (commented):
I've implemented the changes based on your suggestions—you can check out the latest commits! Let me know if everything looks good or if you’d like any further adjustments.

boringNotch/extensions/DataTypes+Extensions.swift
```diff
@@ -42,6 +42,31 @@ extension Date {
...
+        let fullDate = formatter.string(from: value)
+        
+        if let yearRange = fullDate.range(of: "年") {
+            let extracted = String(fullDate[yearRange.upperBound...])
+            return extracted
+        }
+        
+        return fullDate
+    }
+}
```
Alexander5015, line 62: I think we should avoid recreating this every time. I added a suggestion below that makes the formatter static and also should handle the formatting in a cleaner way.


```suggestion
struct LunarDateStyle: FormatStyle {
    private static let formatter: DateFormatter = {
        let f = DateFormatter()
        f.calendar = Calendar(identifier: .chinese)
        f.locale = Locale(identifier: "zh_CN")
        f.dateStyle = .long
        f.dateFormat = "MMMMd"
        return f
    }()

    func format(_ value: Date) -> String {
        Self.formatter.string(from: value)
    }
}
```

boringNotch/components/Calendar/BoringCalendar.swift
Alexander5015, line 1: This view should be refactored. It shouldn't have the hidden view for sizing in a ZStack and it should probably use the `.headline` font for lunar to keep its characters a similar size to the year.

Alexander5015 (changes requested):

Alexander5015:
Closing since this is stale, feel free to open a new PR if you still want to implement this.

PR #1078, opened 2026-03-02
Add Sparkle update channels (stable/beta/nightly) with nightly pipeline

notluquis:
## Summary
This PR adds first-class Sparkle update channel support and aligns CI/CD publishing with channel-based updates.

### App changes
- Add update channel selector in **Settings → About → Software updates**
- Channels available:
  - `Stable`
  - `Beta`
  - `Dev (Nightly)`
- Configure Sparkle updater delegate to use:
  - Dynamic feed URL (`feedURLString(for:)`)
  - Allowed channels (`allowedChannels(for:)`)

### Workflow changes
- Add new workflow: `.github/workflows/nightly.yml`
  - Runs on nightly on `dev`
  - Builds signed DMG via reusable build workflow
  - Publishes immutable nightly releases
  - Generates signed branch appcast
  - Commits appcast pointer updates back to the corresponding branch
- Update reusable build workflow with `commit_version_changes` input to support nightly without committing version bumps
- Update release workflow build number generation to monotonic cross-channel number

## Why
- Enable users/testers to explicitly choose between stable, beta, and nightly tracks
- Keep Sparkle channel behavior consistent with release metadata
- Avoid version-order issues when switching channels

## Notes
- Beta continues to come from the release workflow using Sparkle `beta` channel metadata.


notluquis:
Copilot suggestions already addressed.

CONTRIBUTING.md
```diff
@@ -92,6 +92,12 @@ Please submit all translations to [Crowdin](https://crowdin.com/project/boring-n
 
 4. **Be patient**: Reviews take time. Maintainers will get to your PR as soon as they can.
 
+### Release Channels (Maintainers)
```
Alexander5015: Don't think we need this here, its not really related to contributing guidelines

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,169 @@
+name: Nightly Branch Builds
+
+on:
+  push:
+    branches:
+      - main
```
Alexander5015: We can drop the main nightly (at least for now), as main only gets code changes that are released into stable

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,169 @@
+name: Nightly Branch Builds
+
+on:
+  push:
```
Alexander5015: I don't think nightly builds should be on every push, this should be a job that runs once a day

boringNotch/components/Settings/SoftwareUpdater.swift
```diff
@@ -47,21 +59,36 @@ struct UpdaterSettingsView: View {
...
     
     var body: some View {
-        Section {
+        Section(
+            header: HStack {
+                Text("Software updates")
+            },
+            footer: Text(
+                NSLocalizedString(
+                    "Stable and Beta come from official releases. Main and Dev use nightly builds from those branches.",
```
Alexander5015: I am not sure about making nightly builds available here. Especially without any testing, these nightly builds will often be completely broken, so I feel like a seprating this might make more sense, but I'm not sure yet

Alexander5015 (changes requested):
I haven't had a chance to fully review the implementation yet, but I left some comments. I was also wondering what the testing process for this looked like?

notluquis:
Hey! Addressed all the feedback:

- CONTRIBUTING.md: Removed the Release Channels section

- nightly.yml: Dropped main from the trigger, switched from push-based to a daily cron (3am UTC). Also fixed the metadata job to properly checkout dev and read the SHA/commit message from git (since github.event.head_commit is only available on push events)

- UpdateChannel: Removed the Main (Nightly) option from the picker since there's no longer a nightly pipeline for it. Kept Dev (Nightly) for now

As for testing — the appcast-dev.xml that Sparkle polls doesn't exist yet in the repo. It gets generated and committed by the workflow on the first nightly run after merge. So the update channel can only really be tested once this is merged into dev and the cron runs for the first time.

README.md
```diff
@@ -102,6 +102,18 @@ brew install --cask TheBoredTeam/boring-notch/boring-notch --no-quarantine
 - Use the controls to manage your music like a rockstar.
 - Click the star in your menu bar to customize your notch to your heart's content.
 
+## Update Channels
+
+Boring Notch supports multiple update channels in **Settings → About → Software updates**:
+
+- **Stable**: Official public releases.
+- **Beta**: Prerelease builds from the release pipeline.
+- **Main (Nightly)**: Automated nightly builds from the `main` branch.
```
Alexander5015: Remove from docs
```suggestion
```

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,169 @@
...
+        run: |
+          set -euo pipefail
+          BRANCH_NAME="dev"
+          COMMIT_SHA="$(git rev-parse HEAD)"
+          SHORT_SHA="${COMMIT_SHA::7}"
+          COMMIT_MESSAGE="$(git log -1 --pretty=%s)"
+          TAG="nightly-${BRANCH_NAME}-${SHORT_SHA}"
+          APPCAST_FILE="appcast-${BRANCH_NAME}.xml"
+          BUILD_NUMBER="$(date -u +%Y%m%d%H%M%S)$(printf '%06d%02d' "${GITHUB_RUN_NUMBER:-0}" "${GITHUB_RUN_ATTEMPT:-0}")"
+          ASSET_NAME="boringNotch-${BRANCH_NAME}-${SHORT_SHA}.dmg"
```
Alexander5015: This will break if a new nightly build runs on the same commit. The tag and asset name would stay the same, but the build number would change, producing a new `.dmg` with a different signature. Because the immutable release step, the old DMG would remain instead of being replaced. The appcast would then be updated based on the signature of thenew DMG, causing the nightly release to break.

boringNotch/components/Settings/SoftwareUpdater.swift
```diff
@@ -47,21 +59,36 @@ struct UpdaterSettingsView: View {
...
+                )
+            )
+        ) {
+            Picker(
+                NSLocalizedString("Update channel", comment: "Software updates channel picker label"),
+                selection: $updateChannel
+            ) {
+                ForEach(UpdateChannel.allCases) { channel in
+                    Text(channel.title).tag(channel)
+                }
```
Alexander5015, line 70: When the picker changes channels, don't we need to tell Sparkle that the channel has changed so that it starts fetching from the new channel? I think `resetUpdateCycle()` does this

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,169 @@
...
+          } >> "$GITHUB_OUTPUT"
+
+  build:
+    name: Build nightly DMG
+    needs: [prepare]
+    uses: ./.github/workflows/build_reusable.yml
+    with:
+      head_ref: ${{ needs.prepare.outputs.branch_name }}
+      version: ""
+      build_number: ${{ needs.prepare.outputs.build_number }}
```
Alexander5015, line 109: Although unlikely, it is possible that the branch head will have moved in between steps. It would be good to have a consistent commit from the step above to ensure this will never happen. There may be other steps later as well that use the branch head so pin to a specific commit at the beginning of the workflow and use that throughout

Alexander5015 (dismissed):
I left some feedback. Testing doesn’t need to be done in this repo. It can be done in another repo, such as your fork.

notluquis:
hey, do you any help from my side?

Alexander5015:
I should be good, just hardening and polishing, although if you want to review my final version I would appreciate that. If you are down, I can ping you to review once I'm done.

notluquis:
> I should be good, just hardening and polishing, although if you want to review my final version I would appreciate that. If you are down, I can ping you to review once I'm done.

Okey! Ping me when you are done.

Alexander5015:
@notluquis it should be ready for review. I moved the Swift code out into #1562 so it could merge into `dev` separately from the workflow changes which need to be made on `main`.

Issue #1062, opened 2026-02-27
[FEATURE] Adjustable opening animation speed


ntcho:
### Is your feature request related to a problem?

I really appreciate that you guys added the opening delay option. It'd be great to have the opening animation faster.

Here is something similar for the macOS dock: https://apple.stackexchange.com/questions/33600/how-can-i-make-auto-hide-show-for-the-dock-faster

### Describe the solution you'd like

Another slider right below 'Hover delay` slider would be amazing! The slider can be between 0.00s - 1.00s (in case someone wants to make it longer).

I think 0.01s step would be better than 0.1s step (like the hover delay) for this slider.

### Willingness to contribute

- [ ] Yes, I can write the code for this feature
- [x] Yes, I can help with design/mockups

### Pre-submission checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have provided sufficient detail for the team to understand the request

Amorfx:
Hey ! I just implemented this ! You can now control the opening animation duration via a slider in Settings > General > Notch behavior.

The slider ranges from 0.10s (near-instant) to 1.00s (slow), with the current default of 0.42s preserved so nothing changes out of the box. It applies to all opening modes (hover, tap, and gesture).

Feel free to check out [the PR](https://github.com/TheBoredTeam/boring.notch/pull/1079) and let me know if that covers your use case !

Amorfx:
Since the PR has been merged, I imagine we can close the issue?

ntcho:
Resolved with #1079! Thanks @Amorfx

PR #1079, opened 2026-03-02
feat(settings): add opening animation speed slider (#1062)

Amorfx:
## Summary

Add a slider in Settings > General > Notch behavior to control the opening animation duration of the notch.

Related Issue : #1062

## Changes

- Add a new `openingAnimationResponse` key in `Defaults.Keys` (default `0.42`, preserving current behavior)
- Turn `StandardAnimations.open` into a computed property that reads the user setting
- Add an "Opening animation duration" slider (range 0.10s–1.00s) in the "Notch behavior" section, outside the hover-only conditional block so it applies to all opening modes (hover, tap, gesture)

## Screenshots / Videos
https://github.com/user-attachments/assets/bed8c74a-113a-448e-8c47-ec929c6c3d5a

image

 ## Test plan

  - [ ] Build and launch the app
  - [ ] Verify the slider is visible in Settings > General > Notch behavior
  - [ ] Test low value (0.10): near-instant opening
  - [ ] Test high value (1.00): slow opening
  - [ ] Test default value (0.42): identical to current behavior
  - [ ] Verify the close animation is not affected

### DIFF
--- a/boringNotch/animations/drop.swift
+++ b/boringNotch/animations/drop.swift
@@ -18,7 +18,7 @@ enum StandardAnimations {
     /// Spring animation for opening the notch
     static var open: Animation {
         guard Defaults[.enableOpeningAnimation] else {
-            return Animation.spring(response: 0, dampingFraction: 0.8, blendDuration: 0)
+            return Animation.linear(duration: 0)
         }
         return Animation.spring(response: 0.42 / Defaults[.animationSpeedMultiplier], dampingFraction: 0.8, blendDuration: 0)
     }
@@ -26,7 +26,7 @@ enum StandardAnimations {
     /// Spring animation for closing the notch
     static var close: Animation {
         guard Defaults[.enableOpeningAnimation] else {
-            return Animation.spring(response: 0, dampingFraction: 1.0, blendDuration: 0)
+            return Animation.linear(duration: 0)
         }
         return Animation.spring(response: 0.45 / Defaults[.animationSpeedMultiplier], dampingFraction: 1.0, blendDuration: 0)
     }
--- a/boringNotch/Localizable.xcstrings
+++ b/boringNotch/Localizable.xcstrings
@@ -3218,6 +3218,9 @@
     "Brightness" : {
 
     },
+    "Brightness Source" : {
+
+    },
     "Calendar" : {
       "localizations" : {
         "ar" : {
@@ -7431,6 +7434,9 @@
         }
       }
     },
+    "Enable gradient" : {
+
+    },
     "Enable haptic feedback" : {
       "localizations" : {
         "ar" : {
@@ -10349,6 +10355,9 @@
       }
     },
     "Interaction" : {
+
+    },
+    "Keyboard Source" : {
 
     },
     "Launch at login" : {
@@ -14075,6 +14084,9 @@
         }
       }
     },
+    "Option (⌥) Key Behavior" : {
+
+    },
     "Option key behaviour" : {
       "extractionState" : "stale",
       "localizations" : {
@@ -15529,6 +15541,9 @@
           }
         }
       }
+    },
+    "Replace System OSD" : {
+
     },
     "Request Accessibility" : {
       "extractionState" : "stale",
@@ -17555,6 +17570,9 @@
         }
       }
     },
+    "Show OSD in open notch" : {
+
+    },
     "Show percentage" : {
       "extractionState" : "stale",
       "localizations" : {
@@ -17655,6 +17673,12 @@
           }
         }
       }
+    },
+    "Show percentage (closed notch)" : {
+
+    },
+    "Show percentage (open notch)" : {
+
     },
     "Show power status icons" : {
       "localizations" : {
@@ -17955,6 +17979,9 @@
           }
         }
       }
+    },
+    "Show shadow" : {
+
     },
     "Show sneak peek on playback changes" : {
       "localizations" : {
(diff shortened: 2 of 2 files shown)


ntcho:
Thanks a ton for this! This looks to be exactly what I'm looking for.

Seems like the closing animation is also hardcoded, maybe we can get that into a slider as well? Although having 3 slider might make the settings UI a bit complex. It can be hidden in "Advanced settings" dropdown or something similar.

Amorfx:
> Thanks a ton for this! This looks to be exactly what I'm looking for.
> 
> Seems like the closing animation is also hardcoded, maybe we can get that into a slider as well? Although having 3 slider might make the settings UI a bit complex. It can be hidden in "Advanced settings" dropdown or something similar.

Thanks for the feedback! Glad this is what you were looking for.

Good call on the closing animation. Rather than adding a second slider (which could clutter the UI), we could use a single slider that controls both. The default values are slightly different (0.42 for open, 0.45 for close), but we could maintain that delta so both scale proportionally. That way the user gets one intuitive control without the complexity of managing two separate values. What do you think of that approach? Would be great to get a maintainer's perspective on this as well before moving forward.

ntcho:
I like the idea of having 1 slider to manage both values. But scaling them proportionally can lead to unexpected behavior in the future, where the settings say 0.3s but the actual animation is 0.31s.

Or how about "Animation duration multiplier" slider? It can go from 0.00 to 2.00 (default being 1.00), and it will just multiply the animation duration to that value.

- 0.00 would be no animation
- 0.50 would be 2x speed
- 2.00 would be 1/2x speed

This way users have a clear default to go back to, and it can manage both values in an intuitive way.

This settings sound a bit too technical, so I'm open for ideas as well. Keeping it as 0.00s is also pretty intuitive.

Alexander5015:
A speed multiplier feels more natural than specifying animation time in seconds as it maps directly to how something feels. In my opinion, it’s often more natural to think, 'I want this twice as fast,' rather than thinking about a specific value like 0.3s. And as @ntcho noted, having a default users can easily reset to is a nice bonus.

Amorfx:
Good call on the multiplier approach, makes way more sense than exposing raw duration values.

Its updated. The slider now controls a speed multiplier (default 1.0x) that applies to both open and close animations. Range goes from 0.0x (instant snap) to 2.0x (twice as slow).

image

Let me know what you think!

ntcho:
Thanks for the feedback and such quick turnaround for this.

I love this approach, but I'm a bit concerned concerned the 'Animation duration multiplier' is a bit counter-intuitive for non-technical users, as 0.5x for 2x speed feels backwards.

What do we think about the following logic?

- 0: Disabled (Speed is zero)
- 0.5x: Slow (half the speed)
- 1.0x: Normal (default)
- 2.0x: Fast (twice as fast)

This aligns better with how people perceive speed settings in other apps.

Amorfx:
Yes, that seems like a much better idea. I'll implement it accordingly.

Alexander5015:
I generally agree with that, but I don't like using 0x to disable because:
1. It acts like ∞ rather than 0 from a mathematical sense
2. Leads to incorrect slider behaviour imo between 0 and whatever the next step is, where if a person tries to slow down the speed as much as possible, they end up speeding it up to the maximum speed instead.

A true 0x speed on the animation would mean the animation doesn't complete as the multiplier should be inversely proportional to the animation time. The options I can think of are:

1. ∞ as the upper bound (weird imo)
2. Super fast animation is treated as good enough (idk what our upper bound on animation speed would be, but it would be sufficiently fast as to not need a no animation option)
3. Just make it a separate toggle, enable {some term for opening and closing notch animations}, with a slider that appears below for speed when enabled.

Amorfx:
Personally, I would lean towards solution 3. This is a bit more consistent, especially with certain macOS settings that have this pattern.

Amorfx:
Just pushed the change, went with option 3. Added a "Notch animation" toggle that controls both open and close. When disabled, animations snap instantly. The speed slider (0.1x–2.0x) only shows up when the toggle is on, so no more weird 0x edge case.

image
image

boringNotch/animations/drop.swift
```diff
@@ -15,10 +16,20 @@ enum StandardAnimations {
     static let interactive = Animation.interactiveSpring(response: 0.38, dampingFraction: 0.8, blendDuration: 0)
     
     /// Spring animation for opening the notch
-    static let open = Animation.spring(response: 0.42, dampingFraction: 0.8, blendDuration: 0)
-    
+    static var open: Animation {
+        guard Defaults[.enableOpeningAnimation] else {
+            return Animation.spring(response: 0, dampingFraction: 0.8, blendDuration: 0)
```
Alexander5015: Nitpick: might be better to have `return Animation.linear(duration: 0)` instead for this disabled state beause it is clearer imo.

boringNotch/components/Settings/Views/GeneralSettingsView.swift
```diff
@@ -232,6 +234,17 @@ struct GeneralSettings: View {
                         name: Notification.Name.notchHeightChanged, object: nil)
                 }
             }
+            Toggle("Notch animation", isOn: $enableOpeningAnimation)
+            if enableOpeningAnimation {
+                Slider(value: $animationSpeedMultiplier, in: 0.1...2.0, step: 0.1) {
```
Alexander5015: I think there is a rounding/precision bug here(in the slider range calculation). There are likely better fixes but this one should work

```suggestion
                Slider(value: $animationSpeedMultiplier, in: 0.1...2.01, step: 0.1) {
```

boringNotch/animations/drop.swift
Amorfx: I made the change 😄

ntcho:
Awesome! Agreed on the point it's counter intuitive between 0 (actually infinity) and 0.001.

The switch + slider looks great! Thanks all.

Issue #1077, opened 2026-03-02
[Bug] Cannot select Now playing option after installation


Dysta:
### What happened? What are the steps to reproduce the issue?

hello,

Thank you for the amazing tool. It seems that there is a missing option from Media tab in the settings.

When doing the onboarding, we can select this option but in the scrollable menu

<img width="464" height="145" alt="Image" src="https://github.com/user-attachments/assets/e59b2673-0721-41b4-8463-c82b241faa5d" />

### What did you expect to happen?

We can select the Now playing option in the menu to see when we play a music from a browser on youtube for example

### Boring Notch Version

v2.7.3

### macOS Version

Tahoe 26.3

### Music Source? (If relevant)

Now Playing

### Music App (If using Now Playing)

Firefox

### Website (If using browser for music)

Youtube

### Relevant log output

```shell

```

### Checks

- [x] I haven't found any duplicates with my issue.

Alexander5015:
Occasionally, the app may fail to register the Now Playing source. If you run into this, restarting the app typically fixes it.
In v2.8, we’re making this process more reliable and adding clearer messaging if the source can’t be found after multiple attempts. 

This may also happen after a macOS update since the source relies on an unofficial API that Apple has removed access to in the past, though we haven’t seen similar issues since.

Dysta:
> Occasionally, the app may fail to register the Now Playing source. If you run into this, restarting the app typically fixes it. In v2.8, we’re making this process more reliable and adding clearer messaging if the source can’t be found after multiple attempts.

Hello, I cannot select Now Playing source from the option, even after a restart, if I understand well your comment.

Alexander5015:
Yes, that is correct. Restarting the app usually fixes this issue.

I will investigate the problem further. For now, please try these steps:

1. Start playing a real song on your device. Or if you were playing a real song, try without it playing.

2. Restart the app.

Why this works: The app only checks for "Now Playing" when it first opens. Having real music playing may be more reliable than the app's automatic test.

Dysta:
> Yes, that is correct. Restarting the app usually fixes this issue.
> 
> I will investigate the problem further. For now, please try these steps:
> 
>     1. Start playing a real song on your device. Or if you were playing a real song, try without it playing.
> 
>     2. Restart the app.
> 
> 
> Why this works: The app only checks for "Now Playing" when it first opens. Having real music playing may be more reliable than the app's automatic test.

[...]

PR #1088, opened 2026-03-07
Feature/auto move application

selfsta:
# feat: add "Move to Applications" prompt

## Overview

PR addresses the issue of users running the app directly from the `~/Downloads` folder. While the upcoming release `.dmg` provides a visual guide, this implementation provides a fallback prompt to help users move the app to the correct location.

## Prompt Preview
Screenshot 2026-03-06 at 5 50 43 PM

### Automatic Path Detection

Added `checkForApplicationsFolder()` to verify the app's current location on launch. It checks both the system-wide `/Applications` folder and the user-specific `~/Applications` folder using `NSHomeDirectory()`.

###  Move Logic

Instead of adding an external dependency, I implemented a private `moveSelfToApplications(appPath:)` method.

* **Dynamic App Retrieval:** The script dynamically fetches the app name from `CFBundleDisplayName` or `CFBundleName` to ensure the "activate" command works regardless of naming changes.
* **AppleScript Integration:** Uses `NSAppleScript` to delegate the move to **Finder**. This is the most lightweight way to handle an active process moving itself without requiring complex shell scripts or third-party libraries.

### Integration Point

The check is triggered within `applicationDidFinishLaunching`, placed after the primary window and detector setup to ensure the app remains responsive during the prompt.

### Safety
The implementation uses `runModal()` to ensure the user makes a choice before the app continues into potentially background-only operations.

## How to Test
1. Compile the app.
2. Move the `.app` bundle to any folder besides `/Applications` .
3. Launch the app.
4. **Expected Result:** An informational alert appears asking to move the app.
5. Click **Move**.
6. **Expected Result:** Finder moves the app to `/Applications`, and the app re-activates from the new location.

---

### Note:
Unsure of why other commits are showing in my PR, all my changes were made to boringNotchApp.swift


Alexander5015:
please rebase yout changes onto `dev`

selfsta:
> please rebase yout changes onto `dev`

Done, do I need to open a new PR?

Alexander5015:
You do not need to make a new PR, but it the git history still not correct(it should not show my commits). I edited it to fix it. If you need to do this in the future, run `git rebase -i <branch you are rebasing onto>`

Alexander5015:
Also, make sure you do not let AI coding tools make changes to `Localizable.xcstrings‎` file. I reverted all of the changes it made.

Alexander5015 (changes requested):
I tested this, but it didn’t behave as I expected. Is the intention for it to open a file picker? If so, that might not add much benefit, since users could likely move the file themselves more easily. When I selected a file, nothing happened.

Another thing to keep in mind is that the app will usually be run from a DMG, which is a read-only volume. That would prevent moving the file from there (I tested from a normal folder).

I might be misunderstanding the approach, but because of this it may make sense to close this PR.

If you’re interested in working on something in a similar area—improving user-facing messaging) there’s an important issue, #1077, that’s currently blocking the next update and hasn’t been picked up yet. If you’d like to take a look, I’d be happy to explain the details in DMs.

PR #1097, opened 2026-03-15
Feature/bible verse of the day

ShehanCes:
### Summary
Adds an optional “verse of the day” that can replace the calendar in the home view. Users enable it in Settings → Appearance → Show Bible verse of the day. When enabled, a book icon in the notch header toggles between the calendar and the verse panel.

### Changes

New Feature: Bible Verse of the Day
I’ve introduced a "Verse of the Day" panel to the home screen. It’s designed to be "invisible" until needed—it only fetches data when the panel is actually visible to the user. This feature uses bible-api.com to show the verse of the day. 

To keep things efficient, the app caches verses by date and uses request deduplication. If the feature is toggled off, no network calls are made at all.

Users can now toggle between their calendar and the daily verse via a pill in the header. If a fetch fails or is still in progress, the panel handles it gracefully with "Loading" or "No verse available" states.

A new toggle, "Show Bible verse of the day," has been added to the Appearance settings.

Screenshot 2026-03-16 at 1 56 25 AM

The Bible verse toggle in the notch header uses the same pill style as the Home and Shelf tab buttons and the other header icons (mirror, settings): capsule background with secondarySystemFill when active and clear when inactive, with a white icon, so it fits the existing notch UI.

image

Screenshot 2026-03-16 at 1 55 34 AM

Build Fix: Resolving the "Phantom" File
Fixed a frustrating build error caused by a messy rebase.

What happened: A duplicate reference to OpenNotchOSD.swift appeared in the project file, leading Xcode to look for it in the wrong folder.

The fix: I’ve cleaned up the project.pbxproj file, removing the ghost reference and ensuring the file is only tracked in its correct location (OSD/Views). No actual source code was changed here—just some project-file housekeeping to get the app building again.


ShehanCes:
I hope I bring a difference to many lives using boring.notch, I decided to create this feature because this deemed very useful to me. Cheers!

Ein-Tim:
I vote for this to not be a native feature but rather one that can be implemented as an extension when extensions support is ready.

PR #1100, opened 2026-03-17
feat: Addition of swipe gesture media controls

Kathode-Negative:
I added a gesture handler that allows the user to swipe horizontally in either direction and correspondingly skip forwards (right) or backwards (left) in the media controls, while a media event is playing and the notch is minimized. 
thus being rather cumbersome when hover to open is enabled, this was a deliberate design choice, since I was not using the hover to open function and didn't bother designing a different user flow.

I also added an animation that plays when skipping. To do this, i modified the notch shape code to allow me to move it horizontally. 

Finally, the functionality can be disabled in the settings with the already present toggle.


Alexander5015:
Please fix the merge conflicts. Also, in the future, always base your changes off of `dev`. I updated the PR target branch for you, but in future PRs, please set this to `dev` and not `main`

Kathode-Negative:
Understood! i merged and made sure my implementation is still functional

README.md
Alexander5015, line 1: Remove README changes, it seems like these are only intended to be on your fork.

boringNotch/components/Settings/Views/GeneralSettingsView.swift
```diff
@@ -27,6 +27,7 @@ struct GeneralSettings: View {
     @Default(.showOnAllDisplays) var showOnAllDisplays
     @Default(.automaticallySwitchDisplay) var automaticallySwitchDisplay
     @Default(.enableGestures) var enableGestures
+    @Default(.skipGestureEnabled) var enableSkipGesture
```
Alexander5015: Nitpick: seems like this variable isn't used, so you can delete this line

boringNotch/ContentView.swift
Alexander5015, line 1: Gesture is attached unconditionally and handler is insufficient which means it can fire unintentionally while the notch is open.

boringNotch/ContentView.swift
```diff
@@ -643,8 +688,101 @@ struct ContentView: View {
             }
         }
     }
+    private func handleLeftGesture(translation: CGFloat, phase: NSEvent.Phase) {
+        guard !vm.isHoveringCalendar && (musicManager.isPlaying) else { return }
```
Alexander5015: Is there a reason why you disabled this when music is paused?

boringNotch/models/Constants.swift
```diff
@@ -183,6 +183,7 @@ extension Defaults.Keys {
     
     // MARK: Gestures
     static let enableGestures = Key<Bool>("enableGestures", default: true)
+    static let skipGestureEnabled = Key<Bool>("skipGestureEnabled", default: true)
```
Alexander5015: I would default to `false` here, as the current default will automatically enable this for everyone currently using the app unless they opt-out.

boringNotch/components/Notch/NotchShape.swift
Alexander5015, line 1: None of these changes are used, discard the diff from this file. 

One way to do this is to run `git fetch origin && git restore --source=origin/dev -- boringNotch/components/Notch/NotchShape.swift`(assuming `origin` is the remote for TheBoredTeam/TheBoredTeam)

boringNotch/ContentView.swift
```diff
@@ -31,6 +31,23 @@ struct ContentView: View {
 
     @State private var haptics: Bool = false
 
+    // Gesture state flags (used by left/right gesture handlers)
+    @State private var leftGestureActive: Bool = false
+    @State private var rightGestureActive: Bool = false
+    @State private var horizontalGestureActive: Bool = false
```
Alexander5015: None of these state variables are ever read, remove them.

boringNotch/ContentView.swift
```diff
@@ -220,8 +253,16 @@ struct ContentView: View {
...
+                        // expand briefly by chinPulseWidth when pulsing so the visual chin grows
+                        .frame(width: computedChinWidth + chinPulseWidth, height: vm.chinHeight)
+                        // align the chin rectangle horizontally with the notch pulse shift
+                        .offset(x: chinPulseShift)
+                        // avoid inheriting unrelated implicit animations; we animate chinPulseWidth explicitly when triggering
+                        .animation(nil, value: chinPulseWidth)
                 }
             }
+
+            // (global overlay removed) Using the background pulse so it renders beneath the notch
```
Alexander5015: What global overlay is removed?

Alexander5015 (changes requested):
General comment: Seems like it skips multiple songs quite often, needs refinement so it only skip one song at a time

boringNotch/ContentView.swift
Kathode-Negative: i thought it made less sense to have it be available when not playing, since there is no feedback when a song is switched if you don't have inline preview enabled

boringNotch/ContentView.swift
Kathode-Negative: that's a note left from development. I had one version, where instead of moving the notch, there was a separate layer on top that animated.

boringNotch/ContentView.swift
Alexander5015: I think skipping songs with the gesture should still work when the music is paused, as this is consistent with how the rest of the app is currently handling skips.

Alexander5015:
There is some support for this now in `dev`. Since this PR is failing to build and stale, I am closing it now, but feel free to open a new PR based on the current version of horizontal gestures. It is functional but the UI/animations are kinda rough so they could use some improvements.

PR #1109, opened 2026-03-19
Add a dock where you can place websites for quick access.

nahel-b:
Option to display a dock to create website shortcuts. Option to configure it: icon size, label, image, etc.

Capture d’écran 2026-03-19 à 22 38 23
Capture d’écran 2026-03-19 à 22 38 10


boringNotch/components/Dock/Views/DockView.swift
```diff
@@ -0,0 +1,65 @@
+//
+//  ShelfItemView.swift
```
Alexander5015:
```suggestion
//  DockView.swift
```

PR #1112, opened 2026-03-23
feat: improve YouTube Music observer persistence and add build instru…

cjuriartec:
## Problem
1. When closing YouTube Music (Pear Desktop), the application observer in Boring Notch was being cancelled, preventing it from detecting when the player was reopened.
2. macOS permissions (Calendar, Camera) were being reset continuously due to ad-hoc code-signing issues during local builds.

## Solution
- Prevent `appStateObserver` from being cancelled in `handleAppTerminated` for YouTube Music.
- Updated the README with clear instructions for a "Clean Release" build (`rm -rf build/`) and local command line signing (`codesign`) to stabilize permissions on macOS.

## Testing
Tested locally with Pear Desktop (YouTube Music), verified that Boring Notch correctly reconnects after closing and reopening the player multiple times without requiring permission re-grants.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,38 @@ brew install --cask TheBoredTeam/boring-notch/boring-notch
 3. **Build and Run**:
     - Click the "Run" button or press `Cmd + R`. Watch the magic unfold!
 
+### 💡 Stable Local Build (Permissions & Reconnection Fix)
+
+If you are building Boring Notch to use it permanently on your Mac, follow these steps to ensure macOS remembers your permissions (Calendar, Camera) and keeps the connection to players like YouTube Music stable.
+
+1. **Clean and Build**:
+   ```bash
+   rm -rf build/
+   xcodebuild -project boringNotch.xcodeproj -scheme boringNotch -configuration Release -derivedDataPath build ONLY_ACTIVE_ARCH=NO CODE_SIGN_IDENTITY=- CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO
+   ```
+
+2. **Install to Applications**:
+   ```bash
+   rm -rf /Applications/boringNotch.app
+   cp -R build/Build/Products/Release/boringNotch.app /Applications/boringNotch.app
+   ```
+
+3. **Establish a Stable Identity (Fix Permissions)**:
+   This ensures macOS links your permissions to the final binary's layout.
+   ```bash
+   sudo codesign --force --deep --sign - /Applications/boringNotch.app
+   ```
+
+4. **Restart**:
+   ```bash
+   pkill -x boringNotch || true
+   open /Applications/boringNotch.app
+   ```
+
+> [!TIP]
+> **Improved YouTube Music Support**: I've modified the `YouTubeMusicController` to keep the application observer alive when the music player terminates. This ensures that Boring Notch automatically reconnects as soon as you reopen YouTube Music (Pear Desktop).
+
+
 ## 🤝 Contributing
 
 We’re all about good vibes and awesome contributions! Read [CONTRIBUTING.md](CONTRIBUTING.md) to learn how you can join the fun!
--- a/README.md
+++ b/README.md
@@ -150,38 +150,6 @@ brew install --cask TheBoredTeam/boring-notch/boring-notch
 3. **Build and Run**:
     - Click the "Run" button or press `Cmd + R`. Watch the magic unfold!
 
-### 💡 Stable Local Build (Permissions & Reconnection Fix)
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -150,6 +150,38 @@ brew install --cask TheBoredTeam/boring-notch/boring-notch
...
+   ```
+
+4. **Restart**:
+   ```bash
+   pkill -x boringNotch || true
+   open /Applications/boringNotch.app
+   ```
+
+> [!TIP]
+> **Improved YouTube Music Support**: I've modified the `YouTubeMusicController` to keep the application observer alive when the music player terminates. This ensures that Boring Notch automatically reconnects as soon as you reopen YouTube Music (Pear Desktop).
```
Alexander5015: Remove this

README.md
```diff
@@ -150,6 +150,38 @@ brew install --cask TheBoredTeam/boring-notch/boring-notch
...
     - Click the "Run" button or press `Cmd + R`. Watch the magic unfold!
 
+### 💡 Stable Local Build (Permissions & Reconnection Fix)
+
+If you are building Boring Notch to use it permanently on your Mac, follow these steps to ensure macOS remembers your permissions (Calendar, Camera) and keeps the connection to players like YouTube Music stable.
+
+1. **Clean and Build**:
+   ```bash
+   rm -rf build/
+   xcodebuild -project boringNotch.xcodeproj -scheme boringNotch -configuration Release -derivedDataPath build ONLY_ACTIVE_ARCH=NO CODE_SIGN_IDENTITY=- CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO
```
Alexander5015: Why build a universal binary if it is going to only be used locally? I would remove `ONLY_ACTIVE_ARCH=NO`

README.md
```diff
@@ -150,6 +150,38 @@ brew install --cask TheBoredTeam/boring-notch/boring-notch
...
+   ```
+
+2. **Install to Applications**:
+   ```bash
+   rm -rf /Applications/boringNotch.app
+   cp -R build/Build/Products/Release/boringNotch.app /Applications/boringNotch.app
+   ```
+
+3. **Establish a Stable Identity (Fix Permissions)**:
```
Alexander5015: Why not sign the initial build? This is the recommended way to handle this, and would be simpler as well.

boringNotch/MediaControllers/YouTube Music Controller/YouTubeMusicController.swift
```diff
@@ -185,14 +185,16 @@ final class YouTubeMusicController: MediaControllerProtocol {
...
+            // NOTE: Do NOT cancel appStateObserver here.
+            // It must stay alive to detect when the app launches again.
         }
         
         Task {
             await webSocketClient?.disconnect()
             webSocketClient = nil
         }
         
+        reconnectDelay = configuration.reconnectDelay.lowerBound
```
Alexander5015, line 197: Why is this necessary?

Alexander5015 (changes requested):

README.md
cjuriartec: Okay, I'll do it.

README.md
cjuriartec: Good point, that makes sense. I’ll remove ONLY_ACTIVE_ARCH=NO since this is only for local use.

README.md
cjuriartec: That makes sense, thanks for pointing it out. I’ll switch to signing the build from the start instead of doing it afterward.

boringNotch/MediaControllers/YouTube Music Controller/YouTubeMusicController.swift
cjuriartec: Good question 🙂

We reset it so each reconnection starts from a fast retry.
Since the delay grows with each failure (backoff), if we don’t reset it, a future disconnect could wait a long time ....(e.g. 16s) before even trying again even if it was just a brief drop.

Resetting it to lowerBound **ensures** the next reconnect attempt is immediate and responsive again.

Alexander5015 (approved):
Approving this because it fixes the original bug, but it needs some cleanup/improvements on handling reconnection properly which I am doing on a separate branch. Since it still treats an intentional app quit like a transient socket failure, it can schedule reconnect work even though the app was deliberately closed. It also leaves reconnect sleeps unmanaged, so a pending retry can survive app termination and race with the new initialization path.

Alexander5015:
Also, if you still want to write the documentation for local builds, feel free to open a new PR for that

cjuriartec:
Thank you for the review and approval!  
I’m glad the fix solved the original bug.  
I appreciate your notes on reconnection handling and I’ll keep them in mind.

Alexander5015:
It wasn't really a flaw in your PR, it was an issue in the existing code that wasn't solved by `reconnectDelay = configuration.reconnectDelay.lowerBound`

PR #1145, opened 2026-04-12
Add Apple Intelligence PDF summary and chat

Jesuiskoriel:
## Summary

This PR adds Apple Intelligence PDF summarization to the Shelf.

### Changes
- Adds a new Apple Intelligence PDF drop tile in the Shelf
- Supports dropping PDFs to generate an on-device summary
- Adds an integrated summary panel inside the notch
- Adds an Ask/chat UI to ask follow-up questions about the summarized PDF
- Adds automatic scrolling to the chat when a new question or answer appears
- Adds loading/progress UI with glow styling
- Cleans up drop state when dismissing the summary so the notch does not stay open as if a drop were still active

## Validation

- Built successfully in Xcode
- Tested PDF drop into the Apple Intelligence tile
- Tested summary display inside the notch
- Tested Ask/chat input and response UI
- Tested dismissing the summary after a drop


Ein-Tim:
I'm voting for this not being a core feature but rather something that can be implemented as an extension once extension support is ready.

Jesuiskoriel:
Yeah, that would be cool, honestly. I did this for fun and to test Swift a bit; I'd never done it before. If you have any feedback, feel free to share. Thanks a lot!

PR #1171, opened 2026-04-15
feat: Implements a real time waveform visualizer using Apple Accelerate for FFT

srimanachanta:
This PR implements a FFT (Fast Fourier Transform) backend for the waveform visualizer widget in the notch that models the actual audio values instead of synthetic (random) data as the visualizer currently does. This works in real time using Apple's Accelerate framework which works in-place and is hardware accelerated.

It looks like this:

https://github.com/user-attachments/assets/4e7195f1-c52d-44b6-a470-59780ebb0766

Current metrics show the following resource usage on my machine. The power usage metrics are the same:

Randomized Waveform:
- 70MB Memory
- 0.4-2% CPU Utilization

Realtime Waveform:
- 75MB Memory
- 1-1.6% CPU Utilization

This PR also fixed an issue with album covers being clipped on my machine (maybe different notch widths). If this needs to be moved to a different PR or implemented differently, please let me know!

### DIFF
--- a/boringNotch.xcodeproj/project.pbxproj
+++ b/boringNotch.xcodeproj/project.pbxproj
@@ -156,6 +156,7 @@
 		B1D6FD432C6603730015F173 /* SoftwareUpdater.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1D6FD422C6603730015F173 /* SoftwareUpdater.swift */; };
 		B1F0A0022E60000100000001 /* BrightnessManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1F0A0012E60000100000001 /* BrightnessManager.swift */; };
 		B1FEB4992C7686630066EBBC /* PanGesture.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1FEB4982C7686630066EBBC /* PanGesture.swift */; };
+		F1F2A0A100000000000000F1 /* AudioCaptureManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = F1F2A0A200000000000000F2 /* AudioCaptureManager.swift */; };
 		F38DE6482D8243E7008B5C6D /* BatteryActivityManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = F38DE6472D8243E2008B5C6D /* BatteryActivityManager.swift */; };
 /* End PBXBuildFile section */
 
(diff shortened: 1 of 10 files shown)


Alexander5015:
They can stay as one PR

boringNotch.xcodeproj/project.pbxproj
```diff
@@ -1396,11 +1420,17 @@
 					"$(inherited)",
 					"@executable_path/../Frameworks",
 				);
-				MACOSX_DEPLOYMENT_TARGET = 14.0;
+				MACOSX_DEPLOYMENT_TARGET = 14.2;
```
Alexander5015, line 1423: Can we gate the feature instead of bumping the deployment target?

boringNotch/managers/AudioCaptureManager.swift
```diff
@@ -0,0 +1,479 @@
...
+        var ranges: [Range<Int>] = []
+        var pinks: [Float] = []
+        for i in 0..<Self.barCount {
+            let startHz = exp(logMin + (logMax - logMin) * Double(i) / Double(Self.barCount))
+            let endHz = exp(logMin + (logMax - logMin) * Double(i + 1) / Double(Self.barCount))
+            let startBin = max(1, Int((startHz / nyquist) * Double(halfN)))
+            let endBin = max(startBin + 1, min(halfN, Int((endHz / nyquist) * Double(halfN))))
+            ranges.append(startBin..<endBin)
+            let centerHz = sqrt(startHz * endHz)
+            pinks.append(Float(3.0 * log2(centerHz / Self.referenceHz)))
```
Alexander5015, line 433:
```suggestion
            pinks.append(Float(Self.pinkCompensationSlopePerOctave * log2(centerHz / Self.referenceHz)))
```

boringNotch/managers/AudioCaptureManager.swift
```diff
@@ -0,0 +1,479 @@
...
+    static let shared = AudioCaptureManager()
+
+    static let barCount = 6
+    private static let fftSize = 1024
+    private static let log2n: vDSP_Length = 10
+    private static let ringCapacity = 4096
+    private static let floorDB: Float = -50
+    private static let ceilDB: Float = -25
+    private static let referenceHz: Double = 1000
+
```
Alexander5015, line 29: Add a constant for this magic number (used in another suggestion in this file)
```suggestion
    private static let pinkCompensationSlopePerOctave: Double = 3.0
   
```

Alexander5015 (changes requested):

Alexander5015:
Superseded by PR #1214 (uses this work as the foundation)

PR #1185, opened 2026-04-19
feat(media): implement customizable live activity visibility and timeout logic

yukinoaruu:
- Customizable Visibility: Added hideLiveActivityAfterTimeout setting (Settings -> Media) to toggle between temporary (auto-hide) and persistent Live Activity modes during media playback.
- Customizable Timeout: Added liveActivityTimeout setting with a UI Stepper (1-15s) to allow users to define exactly how long the Live Activity remains visible in temporary mode.
- Strict State Enforcement: Refactored ContentView visibility conditions to strictly observe musicManager.isPlaying. The Live Activity now hides immediately on pause, eliminating the previous idle delay.
- UI Redraw Bugfix: Bound hideLiveActivityAfterTimeout via @Default in ContentView to ensure instant UI updates when toggling the setting, removing the need to manually hover the notch to trigger a redraw.
- Consistent Defaults: Updated standard pop-up (sneakPeek and expandingView) timeouts to a consistent 2.0 seconds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,4 +52,6 @@ build/
 *.ipa
 *.xcarchive
 *.dSYM
-
+*MAP.md
+*JOURNAL.md
+*documentation.md
--- a/.github/workflows/base_ref_check_comment.yml
+++ b/.github/workflows/base_ref_check_comment.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       issues: write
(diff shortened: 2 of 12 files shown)


.gitignore
Alexander5015, line 1: Why did you modify the `.gitignore`?

Alexander5015 (changes requested):
Please rebase and target the `dev` branch

Issue #929, opened 2025-12-21
Like button for Spotify don't show up / doesn't work


berkinefeavci:
### What happened? What are the steps to reproduce the issue?

<img width="1154" height="1342" alt="Image" src="https://github.com/user-attachments/assets/7ea25747-69a3-43ec-b988-a1ec1324867a" /> The button for like is invisible and doesnt work when I click that space

### What did you expect to happen?

Make it visible plaese it is one of the needs

### Boring Notch Version

v2.7.3

### macOS Version

25C56

### Music Source? (If relevant)

Spotify

### Music App (If using Now Playing)

Spotify

### Relevant log output

```shell

```

### Checks

- [x] I haven't found any duplicates with my issue.

Kevlar35:
Duplicate of #873

Alexander5015:
It is not possible currently to like songs with Spotify because they don't provide a way for us do this, unlike Apple Music and YouTube Music.

mstreitner:
@Alexander5015 it would be an extra step, but I guess in theory providing the user an option to authorize with spotify will allow to make this feature work via the Spotify API?
https://developer.spotify.com/documentation/web-api/reference/save-tracks-user
only did a real quick research without PoC tho

Alexander5015:
Not really, at least not in a practical sense. While it’s technically possible, using the Spotify API with multiple users requires an official partnership, which comes with requirements we wouldn’t be able to meet. Because of that, users would need to register their own developer credentials and set everything up manually, rather than enabling it with a few clicks.

Some projects on GitHub take this approach, but it seems like the setup is fairly involved and multi-step, making it feasible mainly for more technically savvy users. We could consider adding this in the future, but for now, it isn’t a priority.

PR #1186, opened 2026-04-19
feat: support third-party like/dislike action URLs via distributed notifications

Myseri:
Currently canFavoriteTrack is only true for Apple Music, so the heart button is permanently disabled for all other sources (Spotify, Pandora, etc.) see [#929](https://github.com/TheBoredTeam/boring.notch/issues/929).

Third-party apps can already broadcast distributed notifications with their playback state. This change adds additional functionality to MusicManager to read optional "likeAction" and "dislikeAction" URL strings from those apps. When present, tapping the heart button opens the URL so the source app handles its own rating logic rather than requiring AppleScript.

This does not change any existing Apple Music behavior, and requires no new dependencies.

Apps would need to include these keys in their notification payload:
  "likeAction":    "myapp://like"
  "dislikeAction": "myapp://dislike"


Alexander5015:
Please rebase/cherry pick your changes onto dev

PR #1190, opened 2026-04-19
Resolve source app icons for helper bundle media sources

anuragkej:
Screenshot 2026-04-19 at 5 31 17 PM
## Summary
- resolve source app icons through a shared helper used by both SwiftUI and NSImage paths
- include running application bundle and executable URLs as icon lookup candidates
- normalize nested helper bundle paths back to the outermost `.app` so browser helper processes still show the main app icon

## Why
Some now-playing sources can report helper-process bundle identifiers instead of the primary app bundle. In those cases the notch badge can fall back to the generic app icon even though the real source app is installed and running. This change keeps the existing badge UI intact and only hardens icon resolution so helper-backed browser/media sources resolve to the main app icon more reliably.

## Testing
- `xcodebuild -project boringNotch.xcodeproj -scheme boringNotch -destination 'platform=macOS' build`
- reproduced a Chrome YouTube audio session locally and confirmed the fix path targets Chrome helper bundle layouts under `Google Chrome.app`

## Notes
- no playback state or controller contracts were changed
- no layout, sizing, or interaction behavior was changed; this is a narrow icon-resolution fix
- I do not have a clean screenshot from the debug build because the local environment kept exposing the installed app to the UI automation layer instead of the built branch binary

### DIFF
--- a/boringNotch/helpers/AppIcons.swift
+++ b/boringNotch/helpers/AppIcons.swift
@@ -9,53 +9,95 @@ import SwiftUI
 import AppKit
 
 struct AppIcons {
-    
+
     func getIcon(file path: String) -> NSImage? {
         guard FileManager.default.fileExists(atPath: path)
         else { return nil }
-        
+
         return NSWorkspace.shared.icon(forFile: path)
     }
-    
+
     func getIcon(bundleID: String) -> NSImage? {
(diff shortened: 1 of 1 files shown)


Alexander5015:
I was unable to reproduce this issue. When do you encounter this?

PR #1212, opened 2026-04-25
Improve battery menu with time remaining, time estimate overhaul, and charging state fixes

lasersushi:
## Summary

- Adds a **Time Until Empty** row to the battery popover, sourced from `kIOPSTimeToEmptyKey`, mirroring the existing Time to Full Charge row
- Shows a **"Calculating…"** state for both time estimates while macOS is still computing them (value = `-1`), instead of showing nothing
- Hides the **"Plugged In"** label while actively charging, since the two states are redundant

## Changes

**Time Until Empty**
- Added a new `timeToDischarge` field to `BatteryInfo`, populated from `kIOPSTimeToEmptyKey`
- Wired it through the full event pipeline — `BatteryActivityManager` → `BatteryStatusViewModel` → views
- Displayed in the battery popover using the same `DateComponentsFormatter` as Time to Full Charge

**Calculating… state for both time estimates**
- Both Time Until Empty and Time to Full Charge now show a "Calculating…" label when macOS returns `-1` (still computing), rather than showing nothing at all

**Charging label fixes**
- "Time to Full Charge" now only appears when the battery is actively charging, not just plugged in
- "Plugged In" is hidden while charging since the two states are redundant

## Testing

Tested locally in Xcode across all relevant battery states (on battery, charging, plugged in but not charging, and the Calculating… transitional state). Builds cleanly with no warnings.

## Screenshots:
Screenshot 2026-04-25 at 2 58 25 PM
Screenshot 2026-04-25 at 2 59 51 PM
Screenshot 2026-04-25 at 3 01 23 PM


Alexander5015 (approved):
Looks good!

lasersushi:
Thanks!

PR #1216, opened 2026-04-26
Stop notch clicks from leaking through to the menu bar

AlexandrosAlexiou:
The notch window is a non-activating panel with canBecomeKey = false and a stock NSHostingView. NSView.acceptsFirstMouse(for:) defaults to false, so clicks on SwiftUI buttons inside the panel were not treated as a real first-mouse event and continued propagating through the WindowServer chain. Apps monitoring the menu bar (e.g. Ice) saw the click as a menu-bar click.

Wrap the contentView in a FirstMouseHostingView that overrides acceptsFirstMouse to true, so the click is fully owned by our window.

### DIFF
--- a/boringNotch/boringNotchApp.swift
+++ b/boringNotch/boringNotchApp.swift
@@ -276,7 +276,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             window.disableSkyLight()
         }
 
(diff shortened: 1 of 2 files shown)


AlexandrosAlexiou:
This is an Ice issue apparently, closing this. Sorry for the noise.

PR #1220, opened 2026-04-27
Add Yandex Music integration

anatolykopyl:
Added [Yandex Music](https://music.yandex.ru/download/) as a media source. Relied on the existing Apple Script integrations for examples.

Could not get shuffle, loop, volume and favorites to work, prodded at the Yandex Music app and it did not respond to anything of the above.


Alexander5015:
Hey @anatolykopyl! Thanks for working on this. I know the integration is still a draft, but I wanted mention that if the final Yandex Music integration depends on wrapping the Now Playing controller, I won’t be able to accept the PR. The integration should be able to function independently, like the other music controller implementations.
There are two main reasons for this:
1. If Yandex Music already works through Now Playing, then a thin wrapper around Now Playing doesn’t add much value.
2. Now Playing is considered unstable, so integrations shouldn’t rely on it as their core implementation.
A Yandex Music controller would need to talk to Yandex Music directly (through a REST API or AppleScript for example), rather than depending on Now Playing.

anatolykopyl:
Thanks for the feedback! After some investigation, I found that moving away from Now Playing isn’t feasible. So, I’m closing this PR, at least until a way to control Yandex Music is available.

PR #1223, opened 2026-04-28
Add Timer tab to the notch with countdown and preset controls

nastrazeneca:
Introduces a new Timer tab between Home and Shelf (Shelf remains available when enabled in settings). The tab bar is now always visible when the notch is open. Timer supports 1/5/10/15/30m presets, +/- minute adjustments, start/pause/resume/reset, a circular progress ring, and a smooth numeric countdown using the app's existing visual language.

https://claude.ai/code/session_01TSFyFaQ68k9vA7L5HUizYK


Alexander5015:
Closing because stale

PR #1233, opened 2026-05-01
Detect shelf drags via NSDraggingDestination instead of global event monitor

### LINKED ISSUE #1175: [Bug] Drag to Airdrop feature not working consistently on connected displays

After installing boring notch on my laptop, I plugged it into my dock and closed the lid. On my monitor, I tried to drag a file from my desktop to the airdrop window and it wouldn't go blue as if to accept the file. The notch opened up, but it wouldn't let me drop anything in there. A restart of boring notch fixed the issue. I HAVE been able to recreate the issue – every time I undock and redock my laptop, it requires a restart to work right on my external monitor. 

(SEE ABOVE)

### Boring Notch Version

v2.7.3

### macOS Version

Tahoe 26.4.1

None

_No response_

_No response_

### Relevant log output

```shell

```

### Checks

gz2436:
Drags toward the notch no longer expand the shelf on macOS Tahoe 26.x — the file falls onto whatever window sits below the closed notch. closes #1175.

The current `DragDetector` infers "the user is dragging content" by combining global mouse-event monitors with `NSPasteboard(name: .drag).changeCount` deltas between mouseDown and mouseDragged. On macOS 14+, and consistently on Tahoe, Finder / browsers / Mail use lazy / promised pasteboards that don't bump `changeCount` during the drag, so `isContentDragging` never flips and `onDragEntersNotchRegion` never fires.

This replaces the implementation with a transparent always-on-top `NSPanel` sized to the notch hit-rect that registers as `NSDraggingDestination`. AppKit delivers `draggingEntered:` regardless of pasteboard semantics, so the same `handleDragEntersNotchRegion` callback fires reliably.

The external API on `DragDetector` (`startMonitoring`, `stopMonitoring`, `onDragEntersNotchRegion`, `onDragExitsNotchRegion`) is preserved; only the initializer signature changes from `notchRegion: CGRect` to `screen: NSScreen`, since the panel computes its own hit-rect from `openNotchSize` and the screen frame.

Tested on Mac mini + external display, macOS Tahoe 26.4.1 — drags from Finder, Safari, Chrome (URL + image) and Mail all expand the shelf and the dropped item lands in it. Repro for the original failure follows the path described in #1175. Happy to attach a screen recording if it would help review.

### DIFF
--- a/boringNotch/boringNotchApp.swift
+++ b/boringNotch/boringNotchApp.swift
@@ -223,27 +223,15 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
     private func setupDragDetectorForScreen(_ screen: NSScreen) {
         guard let uuid = screen.displayUUID else { return }
-        
-        let screenFrame = screen.frame
-        let notchHeight = openNotchSize.height
-        let notchWidth = openNotchSize.width
-        
-        // Create notch region at the top-center of the screen where an open notch would occupy
-        let notchRegion = CGRect(
-            x: screenFrame.midX - notchWidth / 2,
-            y: screenFrame.maxY - notchHeight,
-            width: notchWidth,
-            height: notchHeight
-        )
-        
-        let detector = DragDetector(notchRegion: notchRegion)
-        
+
+        let detector = DragDetector(screen: screen)
+
         detector.onDragEntersNotchRegion = { [weak self] in
             Task { @MainActor in
                 self?.handleDragEntersNotchRegion(onScreen: screen)
             }
         }
-        
+
         dragDetectors[uuid] = detector
         detector.startMonitoring()
     }
--- a/boringNotch/observers/DragDetector.swift
+++ b/boringNotch/observers/DragDetector.swift
@@ -5,119 +5,133 @@
 //  Created by Alexander on 2025-11-20.
 //
 
-import Cocoa
-import UniformTypeIdentifiers
+import AppKit
 
 final class DragDetector {
 
-    // MARK: - Callbacks
-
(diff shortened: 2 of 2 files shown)


Alexander5015:
I tested this on macOS 15 and macOS 26, and the extended drag detection no longer appears to work as expected. It seems to only rely on the non-extended drag detection, even when extended drag detection is enabled.

gz2436:
Thanks for testing this on 15 and 26 — that pointed straight at the cause. The hit panel was being shadowed by `BoringNotchSkyLightWindow` because they weren't in the same `CGSSpace`, so the dragging session never reached the new panel.

I've force-pushed an update that:

1. Inserts the hit panel into `NotchSpaceManager.shared.notchSpace` so it's enumerated as a sibling of the notch window. Within the space, `level = .mainMenu + 4` keeps it just above the notch panel's `.mainMenu + 3`.
2. Drops `ignoresMouseEvents = true` and adds a `hitTest` gate on the `.drag` pasteboard, so the panel only claims its rect while a drag is in progress and clicks fall through to the menu bar / windows underneath when nothing is being dragged.

Tested locally on Tahoe 26.4.1 — drag-into-shelf opens correctly, and clicking on title-bar buttons positioned in the hit-rect still works. I don't have macOS 15 or a notch-equipped machine to verify against — would you mind giving it another go when you have a moment?

Alexander5015:
Extended drag detection is still not working on either OS for me

gz2436:
Closing — my mistake. A reinstall resolves the symptom; I should have tested that before opening this. Thanks for the time spent reviewing.

PR #1245, opened 2026-05-11
Add caffeinate button to Notch header to keep computer awake when necessary

lasersushi:
### What it does
Adds a coffee cup icon to the notch header that toggles sleep prevention on and off. When active, the system is kept awake indefinitely using a native IOKit power assertion — the same mechanism as macOS's built-in caffeinate command, but without spawning a subprocess. The icon fills and turns yellow while active, and reverts when toggled off or when the app quits.

### What changed
CaffeinateManager.swift (new) — singleton that wraps IOPMAssertionCreateWithName with kIOPMAssertionTypePreventUserIdleSystemSleep. Holds the assertion ID and releases it cleanly on deinit.
BoringHeader.swift — renders the button when showCaffeinateButton is enabled. Uses cup.and.saucer (white, outline) when off and cup.and.saucer.fill (yellow) when on, with a tooltip that reflects current state.
Constants.swift — adds showCaffeinateButton defaults key, enabled by default.
AppearanceSettingsView.swift — adds a toggle to show/hide the button from settings.
Localizable.xcstrings — adds localization strings for the new UI.

### Testing
✅ Builds cleanly in Xcode with no warnings or errors
✅ IRL functionality test — button toggles correctly, icon and color update as expected, system stays awake while enabled
✅ Verified kernel assertion is created and released via pmset -g assertions | grep -i caffeinate

### Screenshots:
</div>
image
<div>
</div>
image
<div>

### Notes
Uses native IOPMAssertionCreateWithName — no caffeinate subprocess, no shell dependency.
Assertion is properly released on both manual toggle-off and app termination (deinit).
Button is shown by default, user can turn it off in the appearance settings.

### Duration dropdown

Added a duration picker so users can pick how long caffeinate stays on, instead of being limited to indefinite.

### What it does (update)
Clicking the cup icon while it's **off** now opens a dropdown with: 5 minutes, 15 minutes, 30 minutes, 1 hour, 2 hours, and Indefinite. Clicking while it's **on** still disables it immediately, same as before. If a duration is set, caffeinate auto-disables when it elapses, and the tooltip shows the end time (e.g. "Caffeinate: on until 3:45 PM").

### What changed (update)
- **`CaffeinateManager.swift`** — `enable()` now takes an optional `TimeInterval`. When non-nil, schedules a one-shot `Timer` on the main run loop that calls `disable()` when it fires. Added a published `endDate` for tooltip display. Timer is invalidated on manual disable and in `deinit`. `toggle()` was removed since the view is now state-aware.
- **`BoringHeader.swift`** — branches on `caffeine.isActive`: a `Button` when on (disables on click), a `Menu` when off (shows the duration dropdown). Capsule + cup icon factored into a shared `caffeineCapsule` view so both branches look identical.

### Testing (update)
- ✅ Menu opens when off, all six options visible
- ✅ Picking "5 minutes" enables caffeinate and tooltip shows correct end time
- ✅ Timer fires after duration elapses — button returns to off, assertion released (verified via `pmset -g assertions`)
- ✅ Clicking while on disables immediately, invalidating the timer cleanly (no stale fire later)
- ✅ "Indefinite" option behaves identically to previous behavior

Screenshot:
image


lasersushi:
Hey @Alexander5015,
Quick update. After seeing another mac app's caffeinate feature, I thought it would be a good idea to update the pull request and add a dropdown from the caffeinate button where you can select how long you want it on, as well as indefinitely. I updated the pr with some new details about it. I am happy to adjust it if you want any changes. Thanks!

lasersushi:
This is also related to issue #1310

KohliNaman:
love this, would love to keep the caffeine option in my notch

PR #1267, opened 2026-05-20
Add customizable notch border color and width

Yoorz-Truely:
So, I've installed your app and have been loving it, but one thing I couldn't help myself finding is that there was no toggle for a colored border, and having the notch against a black background, it can get hard to distinguish. Since this project is open source (which is amazing), I added an optional colored outline around the notch shape, configurable under **Settings → Appearance → Notch border**. This is my very first PR and if you'll like it, I'd love to have it merged so more people can benefit from this toggle.

## Features

- Off/Custom toggle, 8 color presets, free color picker, and a width slider (0.5–6 pt)
- Border traces left, bottom, and right edges only — no line across the top camera area
- Upper inward shoulder curves shown when closed; expand with the notch when **Corner radius scaling** is ON
- On external monitors (no physical notch): flat top always, unless Corner radius scaling is ON
- All settings persisted via `Defaults` keys — no state lost across launches

## Files changed

- `boringNotch/models/Constants.swift` — three new Defaults keys (`useCustomBorderColor`, `customBorderColorData`, `notchBorderWidth`)
- `boringNotch/extensions/Color+AccentColor.swift` — `Color.effectiveBorder` / `NSColor.effectiveBorder` helpers
- `boringNotch/components/Notch/NotchShape.swift` — `NotchBorderShape` (left/bottom/right open path with animatable `topInset`) and `NotchUpperCurvesShape` (upper shoulder curves)
- `boringNotch/ContentView.swift` — border overlay wired to notch state, `cornerRadiusScaling`, and physical-notch detection via `safeAreaInsets`
- `boringNotch/components/Settings/SettingsView.swift` — Notch border section placed in Appearance (after General); `PresetAccentColor` extracted to top-level so both Appearance and Advanced can share it


CLAUDE.md
```diff
@@ -0,0 +1,75 @@
+# CLAUDE.md
```
Ein-Tim: This should not be in here.

Ein-Tim (changes requested):

CLAUDE.md
Yoorz-Truely: Removed

CLAUDE.md
Ein-Tim, line 1: @Yoorz-Truely You did not actually remove the file but only added a configuration to gitignore. You still have to manually remove this file.

PR #1283, opened 2026-05-28
Add Spotify queue support

thehugapp:
Adde Spotify queue viewing and queue-item playback

https://cap.so/s/xvp2nmw83wpqxg4


Alexander5015:
Unfortunately, I don’t think this will work. We don’t have a Spotify Premium account for the Web API, and even if we did, we’d be limited to 5 users.

Issue #1286, opened 2026-05-29
[Bug] Default calendar app not respected


rantoniuk:
### What happened? What are the steps to reproduce the issue?

I have Fantastical set as my default calendar app. When I click on an event in Boring Notch, the native Calendar app is opened instead of Fantastical.

### What did you expect to happen?

Fantastical should open (as the default calendar app).

### Boring Notch Version

v2.7.3

### macOS Version

26.5

### Music Source? (If relevant)

None

### Relevant log output

```shell

```

### Checks

- [x] I haven't found any duplicates with my issue.

Ein-Tim:
boring.notch only supports the native calendar, if Fantastical synchronises it's events into Apple Calendar, you can still see them in boring.notch. See #1270 for more info.

rantoniuk:
I disagree - this is different scope/use-case. I _do_ have events visible in the Apple Calendar (via synchronised external account from Outlook) and as a result in Boring Notch - that's not the problem here.
The problem is what _App_ is opened when an event is _Clicked_ in the UI of Boring Notch - and that should respect the System setting of the `Default Calendar App`.

Ein-Tim:
Yes, I understand you. But this is not a bug then but a feature request. boring.notch is behaving as intended. Supporting other calendar apps should be an extension feature IMO, as written in the other issue. 

What you are suggesting (supporting deep-links into other calendar apps when the event is clicked) seems out of scope for a core feature, mainly because of the sheer amount of apps boring.notch would need to support (e.g. Outlook, Notion, Fantastical, etc.).

rantoniuk:
Got it, I created the initial implementation - but after reading your comment I will update it to be more maintainable. I'll close this issue and we can continue discussion in the PR itself.

Issue #1270, opened 2026-05-22
[FEATURE] Adding Outlook calendar / reminders


waazdakka:
### Is your feature request related to a problem?

Amazing addon for our Macbook's notch! 

It would be amazing if you could add support for Outlook since I'm not using Apple's calendar and reminders at all

Thanks a lot!

### Describe the solution you'd like

I would like the notch to display my Outlook's calendar and reminders (Or, use the default mail application users set on Settings panel?)

### Willingness to contribute

- [ ] Yes, I can write the code for this feature
- [ ] Yes, I can help with design/mockups

### Pre-submission checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have provided sufficient detail for the team to understand the request

Ein-Tim:
Similar to https://github.com/TheBoredTeam/boring.notch/issues/1001#issuecomment-3825712641, I'd recommend adding Outlook to the Mac Calendar.

waazdakka:
Yeah unfortunately, not happening since my company disable Apple's calendar app :')

Ein-Tim:
In this case I doubt that your company would be happy if you connected your calender to some random third party application. ;) 

Anyway, I think the best idea for such a feature would be an extension ([status of extension support](https://github.com/TheBoredTeam/boring.notch/pull/970#issuecomment-3797144788)).

Alexander5015:
I agree with @Ein-Tim that you should be cautious about any company policies around this. You don't actually need to use the Apple Calendar app, we pull from the system calendars, which can be added in System Settings > Internet Accounts. I'm not sure if this is accessible on managed devices with Apple's Calendar app disabled though.

PR #1288, opened 2026-05-29
support ical event deep-linking to open non-default Calendar apps  

rantoniuk:
### Summary

  Root cause: EventModel.calendarAppURL() always returned an ical://ekevent/... URL. This scheme is Apple Calendar's proprietary internal URL and is hardwired to launch Apple Calendar — no third-party app (including Fantastical) ever registers to handle
  it. The macOS "Default Calendar App" setting has zero effect on ical:// routing.

### Fix (EventModel.swift):

  Before opening an event, the code now queries NSWorkspace.shared.urlForApplication(toOpen: URL(string: "webcal://x")) — which is exactly the mechanism macOS uses to resolve the "Default Calendar App" preference. If the resolved app's bundle ID contains
  "fantastical", it returns a x-fantastical2://show?eventIdentifier=<id> URL instead, which Fantastical does handle. Non-Fantastical installs fall through to the existing ical:// path unchanged.

To extend this for other third-party apps (e.g. BusyCal), can add more bundleId.contains(...) branches in the same block.

Related issues:
- #1286
- #1270


Ein-Tim:
@Alexander5015 would need to take a look at this, but thank you for proposing a possible implementation @rantoniuk!

PR #1304, opened 2026-06-04
Add Wi-Fi connectivity live activity

TheBestVivBoy:
## Summary
Add a small Wi-Fi connectivity live activity that appears when the Wi-Fi path connects or disconnects.

## What changed
- added a lightweight `NWPathMonitor` service for the Wi-Fi interface
- added a small `NetworkStatusViewModel` that suppresses startup noise and only triggers on real state changes
- added a compact closed-notch network activity view and wired it into the existing expanding activity path

## Notes
This first version stays intentionally narrow. It does not read SSIDs, hotspot names, or use any extra entitlements or private APIs. It also avoids claiming "no internet" because `NWPathMonitor` alone does not detect that reliably.

## Validation
- built with `xcodebuild -scheme boringNotch -project boringNotch.xcodeproj -configuration Debug build CODE_SIGNING_ALLOWED=NO`


TheBestVivBoy:
Closing this draft to replace it with a fresh PR after the network activity spacing adjustments.

PR #970, opened 2026-01-12
feat: Add Claude Code session monitoring integration

serrrfirat:
## Summary
- Add Claude Code session monitoring to track active Claude Code terminal sessions
- Add terminal session detection for iTerm2, Terminal.app, and Warp
- Display real-time session stats including token usage, context, and tool activity

## Test plan
- [ ] Open a Claude Code session in iTerm2/Terminal/Warp
- [ ] Verify session is detected and displayed in the notch
- [ ] Check token and context stats update in real-time
- [ ] Test with multiple simultaneous sessions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1306, opened 2026-06-06
feat: add AI Quota tab showing Claude & Codex usage

meloer101:
## Summary
- **Auth failure backoff**: When Claude token expires (401), back off 15 minutes instead of retrying every 2 minutes until rate-limited (429)
- **Rate limit handling**: Parse `Retry-After` header on 429 responses, default 30-min backoff
- **XPC-first credential flow**: Try XPC helper before Keychain fallback to avoid macOS password prompts on unsigned builds
- **Preserve last successful data**: Don't overwrite working quota data with transient errors
- **UI improvements**: Show "Updated X ago" timestamp, use `floor()` for percentage display

## Problem
When the Claude OAuth access token expires, the app was:
1. Fetching with expired token → 401
2. Retrying every 2 minutes with the same bad token
3. Eventually getting rate-limited (429)
4. Only then backing off (30 min)

Now it detects the 401 immediately and backs off with a clear "Token expired. Re-login with CLI." message.

## Test plan
- [ ] Verify Claude quota displays correctly with valid token
- [ ] Verify expired token shows "Token expired" message and backs off
- [ ] Verify rate-limited state shows retry countdown
- [ ] Verify Codex quota unaffected
- [ ] Verify "Updated X ago" timestamp appears on quota cards

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/AI_QUOTA_FEATURE_PLAN.md
+++ b/AI_QUOTA_FEATURE_PLAN.md
@@ -1,0 +1,360 @@
+# Boring.Notch — AI 额度显示功能计划
+
+> **v2** — 整合 Codex 审核反馈，修正 sandbox / 布局 / 生命周期 / 设置入口问题
+
+## 功能概述
+
+在 boring.notch 的 notch 展开视图中新增一个 **AI Quota** 面板，实时显示 Claude (Anthropic) 和 Codex (OpenAI/ChatGPT) 账号的使用率（已用百分比）。
+
+参考项目：[cc-switch](https://github.com/farion1231/cc-switch)
+
+---
+
+## 核心原理（来自 cc-switch 的逆向分析）
+
+### Claude 额度查询
+
+- **凭据来源**：优先从 macOS Keychain（service: `"Claude Code-credentials"`）读取，回退到 `~/.claude/.credentials.json`
+- **JSON 格式**：
+  ```json
+  {"claudeAiOauth": {"accessToken": "...", "expiresAt": ...}}
+  // 或
+  {"claude.ai_oauth": {"accessToken": "...", "expiresAt": ...}}
+  ```
+- **API 端点**：`GET https://api.anthropic.com/api/oauth/usage`
+- **请求头**：
+  - `Authorization: Bearer {token}`
+  - `anthropic-beta: oauth-2025-04-20`
+  - `Accept: application/json`
+- **返回数据**：按窗口维度的 `utilization`（0-100%），已知窗口类型包括：
+  - `five_hour` — 5 小时滑动窗口
+  - `seven_day` — 7 天窗口
+  - `seven_day_opus` — 7 天 Opus 模型窗口
+  - `seven_day_sonnet` — 7 天 Sonnet 模型窗口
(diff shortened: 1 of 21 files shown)


Ein-Tim:
I don't understand what this PR does, the description make it seem like a bug fix (while the title talks about a feature), but this can't be as there is no feature relating to Claude Code in the app right now.

meloer101:
Hi @Ein-Tim, sorry for the confusing description!

This PR adds a **brand new AI Quota tab** to boring.notch — it's a completely new feature, not a bug fix. The branch name and original description were misleading, apologies for that.

## What this PR adds

A new **AI Quota** view accessible from the notch that shows real-time usage for:
- **Claude** (via Anthropic's OAuth API) — displays 5h and 7d usage windows with progress bars, reset countdown, and last-updated timestamp
- **Codex** (via OpenAI's usage API) — same display format

The feature reads credentials from the local Claude Code CLI keychain and Codex CLI config, so it works out of the box if the user has either tool installed.

## The "fix" commits

The `fix:` commit in the branch addresses a bug discovered during development: when the Claude access token expires (401), the original code would retry every 2 minutes with the same dead token until it got rate-limited (429). The fix adds a proper backoff policy for both auth failures (15 min) and rate limits (30 min / Retry-After header).

Happy to update the PR description or split into separate PRs if that's easier to review. Let me know!

meloer101:
Also wanted to mention — this is actually my first ever PR on GitHub, so I'm still learning the process! I just noticed myself that the branch is named `fix/` when it should really be `feat/`. I wasn't paying close enough attention when setting it up. Thanks for taking the time to look at it and leave a comment, really appreciate it! 🙏

meloer101:
One known issue I should flag: before the quota data loads successfully, the app prompts for the macOS keychain password multiple times. This happens because the app falls back to reading the Claude credentials directly from the system keychain (via `SecItemCopyMatching`), which triggers the macOS authorization dialog on unsigned/locally-signed builds.

We have an XPC helper that tries to read credentials from the filesystem first to avoid this, but it currently falls back to the keychain path too often. We haven't been able to fully resolve this yet and would appreciate any guidance from someone more familiar with the project's XPC/keychain setup. Happy to iterate on this if you have suggestions!

Ein-Tim:
Ah, I see. Don't worry about that. This PR then (at least partly) duplicates the work done in #970. Please take a look at #970, also in regards to the discussion whether this should be a core feature or rather an extension one.

tk1475:
Hi @meloer101, I've been building something similar on my fork (Claude and Codex usage on the notch, using the same Keychain sign-in and usage endpoints), so I'd rather help get this one landed than open another PR.

The branch has conflicts with `dev` now. If it helps, I can rebase it and resolve them, or send the fixes as a PR against your branch. One thing I'd suggest adding from my side: never refreshing the Claude token from the app, since a refresh rotates it and signs Claude Code out.

@Ein-Tim, if #970 is the preferred direction for Claude usage, I'm happy to help there instead.

PR #1330, opened 2026-06-17
Add optional week-aligned calendar view with selectable week start (#538)

### LINKED ISSUE #538: [FEATURE] Allow us to select which day the week starts in the mini calendar
**Is your feature request related to a problem? Please describe.**
I like to have my calendars set to start on Monday for each week. I have this set up in the default Mac Calendar app. on the Boring notch, it starts on...Saturday? Like today is currently Monday and the mini calendar is showing me Sat-Fri. 

**Describe the solution you'd like**
If the Boring Notch doesn't take the calendar settings we set on our machines system-wide...then can we change the setting within the notch specifically?

**Additional context**
Thank you for any clarification you can provide! 

**Checks**

TemujinCalidius:
## Summary

Adds an **opt-in weekly view** for the mini calendar that shows one week at a time, aligned to a user-chosen first weekday — addressing the request in #538 to control which day the week starts on.

The existing scrolling **day dial remains the default**, so current behaviour is unchanged unless the user opts in.

### What's new
- **Settings → Calendar → "Weekly view"** toggle. When on, the calendar shows a fixed 7-day strip starting on the chosen weekday, with `‹ ›` chevrons; the scroll wheel and chevrons page a whole week.
- **"Week starts on"** picker (System default / Sunday / Monday), shown only when Weekly view is enabled.

### Implementation notes
- New `WeekStartDay` enum (`firstWeekday` resolved via `Calendar.firstWeekday`; `system` follows the OS) and two defaults: `calendarWeekView` (Bool, default **false**) and `weekStartDay` (default `.system`).
- New `WeekStripPicker` view; `CalendarView` switches between the original dial and the week strip on the toggle. The dial code path is unchanged.
- Weekday symbols use the locale-aware `Calendar.veryShortWeekdaySymbols`; week boundaries use `Calendar.dateInterval(of: .weekOfYear:)` so they honour `firstWeekday` and are DST-safe.
- Option labels are keyed by their English text so untranslated locales fall back to readable English rather than raw string keys.
- While here, tidied the scroll-wheel `NSEvent` local-monitor lifecycle in **both** pickers (store the token, remove it on `.onDisappear`) — the dial previously added a monitor on every appear without removing it.

### Testing
- Builds clean (`xcodebuild`, Debug, macOS).
- Verified on a real notch: default dial unchanged; toggling Weekly view switches live; Sunday/Monday/System re-align the strip; wheel + chevrons page exactly one week; selecting a day updates the event list.

> UI change — a short screen recording of the dial ↔ weekly toggle will be added.

Closes #538


TemujinCalidius:
Screen Recording - https://makertube.net/w/nycMp9YcooEpYENK6r8U64

Alexander5015:
Can you modify it so that scrolling still changes the selected date without changing the week alignment until the user scrolls outside of the bounds of the current window, and then it changes? (It would also be good to animate this transition, just for scrolling)

boringNotch/Localizable.xcstrings
```diff
@@ -13727,6 +13727,50 @@
...
+      "comment" : "Week starts on: Sunday",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Sunday"
+          }
+        }
+      }
+    },
```
Alexander5015: Use the built in localization for days, Apple provides these through DateFormatter I believe.

boringNotch/components/Settings/Views/CalendarSettingsView.swift
```diff
@@ -33,6 +35,16 @@ struct CalendarSettings: View {
...
             }
+            Defaults.Toggle(key: .calendarWeekView) {
+                Text("Weekly view")
+            }
+            if calendarWeekView {
+                Picker("Week starts on", selection: $weekStartDay) {
+                    ForEach(WeekStartDay.allCases) { day in
+                        Text(day.localizedString).tag(day)
+                    }
+                }
```
Alexander5015, line 46: I think it would be good to copy the design Apple uses for their calendar app. The default option should be system, with the current system setting in parentheses:

Image

and we may as well let the user pick and day of the week:

Image

Alexander5015 (changes requested):

TemujinCalidius:
Thanks for the review, @Alexander5015! Pushed updates addressing all three points:

1. **Built-in day localization** — the "Week starts on" options now use the system's localized weekday names via `Calendar.weekdaySymbols` (Apple-provided) instead of custom `Monday`/`Sunday` string keys.
2. **Apple Calendar–style picker** — the default **System Setting (X)** option shows the resolved system first weekday in parentheses, and you can now pick **any** day of the week (Sunday–Saturday).
3. **Scroll behaviour** — scrolling now moves the selected date **one day at a time** while keeping the week aligned; once the selection leaves the visible week it shifts by one. It reuses the app's existing `withAnimation` (no new animation): the selection glides between days, and the week flip is a contained `.push` slide clipped to the calendar column.

Updated screen recording: https://makertube.net/w/t2vtaqLEHuuSJNYF68F69p

Ready for another look whenever you have time 🙏

Alexander5015 (approved):

Issue #236, opened 2024-12-06
[FEATURE]Unlock animation


ayan-binish:
**Is your feature request related to a problem? Please describe.**
The feature is NOT related to a problem; it is just a feature request that would be cool.

**Describe the solution you'd like**
When you unlock your Mac there could be an animation as such it is when you unlock your ***iPhone*** with *FaceID* but instead of the *FaceID* animation, there could be a lock to unlock animation.

**Additional context**
Please feel free to add any other context or screenshots about the feature request here.
Make the notch square and add animation for ***seamless*** integration.

**Checks**
- [x] I haven't found any duplicates with my issue.

LennartOnGH:
+1

reartvapps:
I think this would be a cool idea to add. A good reference would be alcove.

TemujinCalidius:
I put together a quick proof-of-concept of what an unlock animation could look like in the notch — a lock that drops just below the notch and morphs open the moment you log back in:

▶️ https://makertube.net/w/6xME4grAeA2QG1wx72JCQL

Note this is a rough **mockup to gauge interest, not a finished design** — the timing, size, and style are all placeholder and easy to tweak.

On doing it *on the actual lock screen*: macOS doesn't expose a public API for a third-party app to draw custom UI on the real lock screen (it's a separate, system-owned secure context), so this fires on the **unlock transition** — it plays on the notch the instant you're back at the desktop, rather than over the password field. (That's arguably the natural moment for an "unlock" flourish anyway.)

Is this the kind of thing people are after? Happy to polish it up or adjust the style if there's interest.

Alexander5015:
@TemujinCalidius we do have the ability to draw custom UI over top of the lock screen, if you are interested in using that ability as a part of your mockup

TemujinCalidius:
I took a run at this and have a working implementation up as a draft PR: #1346

Short version: a closed padlock shows in the notch over the lock screen and morphs open on unlock, with a full privacy lockdown while locked (nothing private shown, non-interactive) and media keys + the audio soundwave still working.

🎥 Demo: https://makertube.net/w/fygAdGQPZ6yoKZuYmiY1dZ

The PR write-up walks through a non-obvious click-swallow bug that SkyLight delegation introduces and how the notch window is swapped to fix it flash-free. Opening as a draft to align on scope first — notably that it uses the private SkyLight API (same as the existing `showOnLockScreen`) and whether it should default on. Feedback very welcome!

PR #1346, opened 2026-06-25
feat: lock-screen unlock animation with privacy lockdown (#236)

TemujinCalidius:
Implements the unlock animation from #236.

## What it does
A closed padlock appears in the notch **over the macOS lock screen** and **morphs open** (`lock.fill` → `lock.open.fill`) the moment the Mac is unlocked. The notch widens horizontally just enough to show it.

🎥 **Demo:** https://makertube.net/w/fygAdGQPZ6yoKZuYmiY1dZ

- **Privacy lockdown while locked** — nothing private is shown (no calendar, track title, album art, or shelf) and the notch is non-interactive (can't be opened by hover / tap / gesture / drag / shortcut).
- **Music still works** — the non-private audio soundwave shows on the right while playing, and hardware media keys (play / skip / stop) keep working.
- **Toggle** — "Lock animation on unlock" in General settings.
- Drawing over the lock screen reuses the **SkyLight** window the existing `showOnLockScreen` feature already uses.

## Why the `boringNotchApp.swift` diff is large (the journey)
Presenting the notch over the lock screen surfaced a non-obvious bug; most of the added code is the fix, and it informs the design:

1. **Click-swallow after unlock.** SkyLight delegation leaves *persistent server-side state* on the window — after unlock its transparent margins swallow desktop clicks under/near the notch. Undelegating, re-homing into the high-level `notchSpace`, and moving back to the active space all failed to clear it. The only reliable cure is **replacing the window with a brand-new, never-delegated one**.
2. **Flash on swap.** A *fresh* `NSHostingView` shows a blank first frame (whole-notch flash); even reparenting flashed because the moved layer's surface re-hosts asynchronously (sub-frame gap); and a fresh view can't match the **chins** — `AudioSpectrum` uses a `CACurrentMediaTime()`-relative keyframe animation (re-phases the waveform) and album art uses a per-`ContentView` `matchedGeometryEffect` namespace.
3. **What works best (this PR).** Reparent the **same** `NSHostingView` onto the pristine window (keeps spectrum phase + album art continuous), bridge the async re-host with a **static snapshot of the live notch placed beneath it** (never blank), inside a passthrough container so margin clicks still pass through. The old window is retired one runloop turn later.

## Known tradeoff
A faint ~1–2px horizontal seam at the very top edge can briefly appear at the swap **only while media is playing**. With no media the notch is closed, so it never shows. Ideas welcome.

## Open questions
- Private SkyLight API — acceptable here (the existing `showOnLockScreen` already uses it)?
- Default-on vs opt-in for the toggle?
- Scope: keep the on-lock-screen padlock, or just the in-notch unlock morph?

Opening as **draft** to align on scope before polishing.

### DIFF
--- a/boringNotch.xcodeproj/project.pbxproj
+++ b/boringNotch.xcodeproj/project.pbxproj
@@ -101,6 +101,7 @@
 		149E0B972C737D00006418B1 /* WebcamManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 149E0B962C737D00006418B1 /* WebcamManager.swift */; };
 		149E0B9A2C737D40006418B1 /* WebcamView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 149E0B992C737D40006418B1 /* WebcamView.swift */; };
 		14A7E5882C64A89C008C1BE9 /* HelloAnimation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 14A7E5872C64A89C008C1BE9 /* HelloAnimation.swift */; };
+		549AADD9DEBF99A16F0E53E0 /* UnlockAnimation.swift in Sources */ = {isa = PBXBuildFile; fileRef = F7FDBCDFF02732D4F40308B3 /* UnlockAnimation.swift */; };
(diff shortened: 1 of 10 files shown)


TemujinCalidius:
@Alexander5015 this is ready for review now — it uses the over-the-lock-screen capability you mentioned. It's synced onto the latest `dev` and the checks are green. Whenever you have a moment, I'd love your take on the SkyLight approach and whether it should default on vs. opt-in. Thanks!

Issue #1333, opened 2026-06-17
[FEATURE] Minimal progress bar under the notch during media playback


LucaClrk:
### Is your feature request related to a problem?

No, it is a minor inconvenience. Currently, when music is playing, there is no quick, ambient way to see how far along a song is just by glancing at the Boring Notch. To check the track's progress, I have to open the full notch to see.

### Describe the solution you'd like

I would like a minimal progress indicator placed directly underneath the Boring Notch when media is active.

Because space around the notch is limited, the solution could be a simple, thin white line. This line would act as a progress bar, filling up from left to right to visually represent how much of the song has played.

### Use cases & user scenarios

-  I am working on a document or coding, and a song comes on that I really like. Instead of interrupting my typing to click around and see how much time is left in the track, I can just glance up at the notch, see the white line is halfway through, and immediately know where I am in the song.
- I'm listening to a live set or a long concept album and want to know if the track is about to end. A quick look at the thin line nearing the right edge of the notch lets me know a transition is coming up, without me needing to open any windows.

### Willingness to contribute

- [x] Yes, I can write the code for this feature
- [ ] Yes, I can help with design/mockups

### Pre-submission checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have provided sufficient detail for the team to understand the request

dstn-idly:
...(claim comment above)...

dstn-idly:
I've implemented this, a thin, opt-in progress bar that traces the notch's bottom edge, color-matched to the playing media, with an adjustable thickness slider. Opening a PR now if no one's already on it.

dstn-idly:
I've implemented this, a thin, opt-in progress bar that traces the notch's bottom edge, color-matched to the playing media, with an adjustable thickness slider. Opening a PR now if no one's already on it.

PR #1350, opened 2026-06-25
Add optional media progress bar around the notch

dstn-idly:
## What

Adds a thin media progress indicator that traces the notch's bottom contour and fills left-to-right as the current track plays, so you can see song progress at a glance without opening the notch.

## Why

Relates to #1333. A minimal, non-intrusive way to track playback progress (e.g. while focused on work, or anticipating track transitions during long albums/sets).

## Demo
White
image
Match Album Color
image
Bar when you use the custom OSD.
image

Video Example

https://github.com/user-attachments/assets/f4bcdd6a-f7c7-4b0c-ad9c-d747d14b0191

Settings
image
Color Dropdown for Settings.
image

## Details

- Draws a tapered ribbon along the notch's bottom edge (`MediaProgressBar` + `TaperedNotchProgress` in `boringNotch/components/Music/MediaProgressBar.swift`), aligned to the live notch corner radii.
- **Tapers up around both bottom corners.** The leading edge stays full thickness mid-song and only tapers as it nears the right corner, so it looks clean during normal playback.
- **Selectable color** via a picker in **Settings → Media**, **Album art / White / Accent**,  mirroring the existing Appearance "Slider color" option (reuses the same `SliderColorEnum`). Album art uses the playing media's average color (the source the audio visualizer uses); Accent uses the app's effective accent.
- **Adjustable thickness** via a slider in **Settings → Media** (1–5 px).
- **Disabled by default** (`showMediaProgressBar`). When enabled, it fades out *before* the notch expands on hover and fades back in once it settles, so it only shows in the idle state.
- Reads existing playback state from `MusicManager` (`estimatedPlaybackPosition`),  no changes to media plumbing.

## Testing

- `xcodebuild -scheme boringNotch ... build` → **BUILD SUCCEEDED**.
- Verified live: bar appears under the notch during playback, the thickness slider and color picker (Album art / White / Accent) update it in real time, it tapers around the corners, and it fades out/in correctly on hover.

## Note for reviewers

The color picker reuses the existing `SliderColorEnum` (Album art / White / Accent) and its already-translated labels rather than introducing a parallel enum, keeping it consistent with the Appearance "Slider color" option. The picker titles ("Progress bar color" / "Progress bar thickness") will be picked up by the string catalog on the next localization sync.

### DIFF
--- a/boringNotch.xcodeproj/project.pbxproj
+++ b/boringNotch.xcodeproj/project.pbxproj
@@ -96,6 +96,7 @@
 		14288E0C2C6F8EC000B9F80C /* AppIcons.swift in Sources */ = {isa = PBXBuildFile; fileRef = 14288E0B2C6F8EC000B9F80C /* AppIcons.swift */; };
 		1443E7F32C609DCE0027C1FC /* matters.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1443E7F22C609DCE0027C1FC /* matters.swift */; };
 		147163982C5D35B70068B555 /* MusicVisualizer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 147163972C5D35B70068B555 /* MusicVisualizer.swift */; };
+		B158C813C0C84091B5F469FA /* MediaProgressBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = 743D0057B1284990B660D980 /* MediaProgressBar.swift */; };
 		1471639A2C5D35FF0068B555 /* MusicManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 147163992C5D35FF0068B555 /* MusicManager.swift */; };
 		1471A8592C6281BD0058408D /* BoringNotchWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1471A8582C6281BD0058408D /* BoringNotchWindow.swift */; };
 		149E0B972C737D00006418B1 /* WebcamManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 149E0B962C737D00006418B1 /* WebcamManager.swift */; };
@@ -279,6 +280,7 @@
 		1443E7F22C609DCE0027C1FC /* matters.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = matters.swift; sourceTree = "<group>"; };
 		1443E7F42C609E650027C1FC /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
 		147163972C5D35B70068B555 /* MusicVisualizer.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MusicVisualizer.swift; sourceTree = "<group>"; };
+		743D0057B1284990B660D980 /* MediaProgressBar.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MediaProgressBar.swift; sourceTree = "<group>"; };
 		147163992C5D35FF0068B555 /* MusicManager.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MusicManager.swift; sourceTree = "<group>"; };
 		1471A8582C6281BD0058408D /* BoringNotchWindow.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BoringNotchWindow.swift; sourceTree = "<group>"; };
 		149E0B962C737D00006418B1 /* WebcamManager.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WebcamManager.swift; sourceTree = "<group>"; };
@@ -792,6 +794,7 @@
 			isa = PBXGroup;
 			children = (
 				147163972C5D35B70068B555 /* MusicVisualizer.swift */,
+				743D0057B1284990B660D980 /* MediaProgressBar.swift */,
 			);
 			path = Music;
 			sourceTree = "<group>";
@@ -995,6 +998,7 @@
 				5955950D2E900ED800C66711 /* ApplicationRelauncher.swift in Sources */,
 				1100292A2E8691B400035A57 /* FileShareView.swift in Sources */,
 				147163982C5D35B70068B555 /* MusicVisualizer.swift in Sources */,
+				B158C813C0C84091B5F469FA /* MediaProgressBar.swift in Sources */,
 				11CC44A22CEE614100C7244B /* BoringViewCoordinator.swift in Sources */,
 				B186543C2C6F49AE000B926A /* ShortcutConstants.swift in Sources */,
 				11A45C792E34E63100CEB175 /* MediaChecker.swift in Sources */,
--- a/boringNotch/ContentView.swift
+++ b/boringNotch/ContentView.swift
@@ -66,22 +66,23 @@ struct ContentView: View {
         return max(0, baseClosedTop * scaleFactor)
     }
 
-    private var currentNotchShape: NotchShape {
+    private var currentBottomCornerRadius: CGFloat {
         // Scale bottom corner radius for closed notch shape when scaling is enabled.
         let baseClosedBottom = cornerRadiusInsets.closed.bottom
-        let bottomCorner: CGFloat
 
         if vm.notchState == .open {
-            bottomCorner = cornerRadiusInsets.opened.bottom
+            return cornerRadiusInsets.opened.bottom
         } else if let scaleFactor = cornerRadiusScaleFactor {
-            bottomCorner = max(0, baseClosedBottom * scaleFactor)
+            return max(0, baseClosedBottom * scaleFactor)
         } else {
-            bottomCorner = displayClosedNotchHeight > 0 ? baseClosedBottom : 0
+            return displayClosedNotchHeight > 0 ? baseClosedBottom : 0
         }
+    }
 
-        return NotchShape(
+    private var currentNotchShape: NotchShape {
+        NotchShape(
             topCornerRadius: topCornerRadius,
-            bottomCornerRadius: bottomCorner
+            bottomCornerRadius: currentBottomCornerRadius
         )
     }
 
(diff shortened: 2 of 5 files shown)


javiBR77:
Good afternoon, and thank you for the great work you're doing. When will the new version with the music bar be released? Or do I need to modify a file after downloading it from here?

javiBR77:
And how do I do it? I can't find those folders.

dstn-idly:
Hey @javiBR77, thank you for the kind words!

Just to explain how this works: this is a pull request, a change I've proposed to the Boring Notch team. It hasn't been merged or released yet, and it's waiting for the maintainers to review it. So there's nothing for you to download or edit right now, that's also why you couldn't find those folders. Although, you can see the files I have edited or changed or added.

Once the maintainers merge it, the music bar would show up in a normal Boring Notch update, you won't have to do anything special, but to activate the feature in the settings menu.

You can also subscribe on the right side of this page and GitHub will notify you when it's merged.

Buttt iif you don't wanna wait or if this feature gets rejected, you can use Xcode and build the app to try it early, you can find it from my "feature/media-progress-bar" branch, but that's totally optional and a bit technical, so waiting for the official release is the easy route.

Although I would be happy to build it for you if you don't have Xcode since its a big download lol. But I won't be responsible for maintaining the app and updating it. So you'll be stuck on that version forever, unless they have an official update with it.

javiBR77:
Thank you so much for your reply. I don't have Xcode; could you please post it for me? I understand the app will remain as it is until they update it, but we don't know if they'll continue updating it. At least being able to use this essential function is a big plus. I use music for work, and it's a very useful feature.

dstn-idly:
Okay, I have it here for you. By the way, don't trust random software from strangers or sources you can't trust... Like me! 

Do your due diligence of checking for viruses, or any harmful actors/actions.  Anyways a virus total scan.

https://www.virustotal.com/gui/file/cbc98984514d201ec517c48226fec6fe7e6ce6e2aa0f6abbb3db7e0c0d24412b/detection

And the zip. You can double check this through VirusTotal.

[boringNotch-musicbar.zip](https://github.com/user-attachments/files/29440902/boringNotch-musicbar.zip)

It might clear all your settings that you done to the BoringNotch, when you install the new one.

To install...

Quit Boring Notch if it's running.
Double-click boringNotch-musicbar.zip to unzip -> you'll get boringNotch.app.
Drag it into your Applications folder, click Replace.
2. Let macOS open it (otherwise it'll say "damaged")

Open Terminal (⌘+Space, type "Terminal", Enter).
Paste this and press Enter:
xattr -cr /Applications/boringNotch.app

3. Open it

Launch Boring Notch. If it asks for Accessibility or Screen Recording permission, allow it (System Settings -> Privacy & Security).
Your settings might not carry over. But to turn it on, go to Settings in BoringNotch -> Media -> "Show progress bar under notch" (thickness + color options are there too).

Have fun!!

dstn-idly:
Oh and let me know how it goes and works!

javiBR77:
Thank you so much for your kindness and help, it works perfectly. I even see there are more options in settings than the original one I had installed. What I can't get to work is the OSD; no matter how many times I enable accessibility permissions, it keeps asking me for them constantly.

dstn-idly:
I'll check it later when I get home. Thanks for letting me know.

dstn-idly:
Hey @javiBR77 , good news, I finally figured out the OSD thing! 

But first, before you install this new one, can you delete the old boringNotch app you downloaded (just drag it to the Trash). I can explain what was actually going on and what I fixed.

So the OSD (that volume/brightness popup) needs a special "Accessibility" permission, and there's a little helper inside the app that handles it. The problem was, the way I built it before, the app wasn't signed properly, kinda like it had a broken ID card. So macOS couldn't verify it and never remembered that you allowed it, which is exactly why it kept asking you for permission over and over forever lol.

What I changed: I rebuilt it the proper way so the app has a valid signature now, so macOS can actually remember the permission this time. I also turned off a security feature that was making the fixed version crash on launch. So now the OSD should work AND stop nagging you a million times.

Here's the fixed one (boringNotch-musicbar-v2.zip). Same steps as before:

[boringNotch-musicbar-v2.zip](https://github.com/user-attachments/files/29441957/boringNotch-musicbar-v2.zip)

Delete the old app first (important!) and zip!!
Unzip this and drag it into your Applications folder
Open Terminal and paste this, then hit Enter: 

xattr -cr /Applications/boringNotch.app

Open it, and when it asks for Accessibility permission, allow it once. This time it'll actually stick.
After that both the OSD and the music bar should work. Thanks for letting me know!

javiBR77:
It works perfectly, thank you so much again for your kindness.

dstn-idly:
no problem!!

javiBR77:
image
image
image
image
Now I have this problem, with the original app, the notch completes the whole bar, however in yours the option to simulate the original notch is missing, as you can see in the photos

dstn-idly:
sorry i'm having a hard time understanding whats the main problem, could you be more in depth?

javiBR77:
Yes, sorry, the problem is the following:
In your modified version the option for Macs that do not have a notch has disappeared, it does not allow you to make the notch as big as the top bar as it does with the original, if you look at the photos, the notch when there is no music is small, you cannot adjust the width.
It's not a criticism, I'm just asking for your help and I appreciate all the help you're giving me
614288756-7dc27102-9510-49d2-8a54-03c7501b8f7e
.

dstn-idly:
Okay, I looked into it, nothing's actually broken.

So what happened: the version you had before was v2.7.3, but I built mine from the newer "dev" code. In the newer version they reorganized the notch size settings and removed the "Match real notch height" option specifically for Macs without a notch, basically since your Mac has no real notch, they figured that option didn't make sense. That's why you see 2 options instead of 3. It's got nothing to do with my music bar stuff, just a change they made in the newer code.

In that same "Notch height on non-notch displays" dropdown, pick "Custom height" instead, a slider pops up and you can slide it to 32 for the original size.. Or pick "Match menu bar height" to match your top bar.


image


let me know if that works.

javiBR77:
Thank you again for taking the time to help me. To be honest, it shocked me at first, but now that I look at it, it looks better this way. Thank you so much for everything.

Alexander5015:
This is exactly right. We removed that setting because it was never intended for external displays. It was carried over when we added external display support.

Aside from my opinion that the setting isn’t useful on an external display, it also can’t work there: it has no way to detect the notch size in pixels, since there isn't one, and so it falls back to the default value.

As for the collapsed look on a notchless display, this is also a new design decision that I thought looked better and replaces the confusing 0 height option of past versions.

javiBR77:
Thank you both so much for your kindness and for taking the time to answer our questions.
When is the next official version expected to be released? Or is it still unknown?

dstn-idly:
Yeah it's no problem and my pleasure to help you.
I'm pretty sure he's working on the update still, theres really no timeline. But you can always be up to date if you learn how to compile the x-code dev branch and update it yourself.

dstn-idly:
I wonder if i should add a gradient color. like how the osd is.
image

javiBR77:
> Sí, no hay problema y es un placer ayudarte.Estoy bastante seguro de que todavía está trabajando en la actualización, realmente no hay una línea de tiempo. Pero siempre puedes estar al día si aprendes a compilar la rama de dev-code x y actualizarla tú mismo.

In September I'm going to start studying microcomputer systems and networks; I suppose I'll start learning then, although I could start practicing.

ericrcwu001:
Really like this feature — using the notch's own contour is a nice touch, and the tapered corners look great. I've been running it locally on top of `dev`.

Flagging a performance issue I hit while profiling something else, since it's easy to miss and easy to fix.

**On an M2 Air (macOS 26.3), with music playing:**

| | Instantaneous CPU |
|---|---|
| Progress bar on | 13–18% |
| Progress bar off | ~0.0% |

A `sample` while playing put nearly all app-code frames inside `MediaProgressBar.swift` — lines 128–145 and 181–193, i.e. the contour sampling and ribbon construction.

**Cause:** `TaperedNotchProgress` declares `animatableData` (line 75), combined with `.animation(.linear(duration: 0.5), value: fraction)` on line 58. That makes SwiftUI interpolate `fraction` across each update and call `path(in:)` on **every display frame**. Each call resamples the contour into ~161 points, then builds cumulative arc-length, centerline, and two offset arrays.

The trouble is how little motion that buys. On a ~3.5 minute track over a ~200pt contour the bar advances **~0.95 px/sec** — so at 60fps each frame moves the leading edge **~0.016 px**. It's roughly 30 full geometry rebuilds per update to produce sub-pixel movement.

**What I changed locally:**

1. **Removed `animatableData` and the implicit `.animation` on `fraction`.** Since `TimelineView` already republishes every 0.5s and that step is well under a pixel, stepping straight to the new value looks the same.

2. **Reduced sampling from 48/64/48 to 12/16/12** (lines 113–115). A ~2pt ribbon can't show that much detail. I checked this geometrically rather than by eye — max deviation of the coarse outline from the fine one, using the real closed-notch radii (`tcr: 6, bcr: 14` from `matters.swift`), across fractions 0.02–1.0 and thicknesses 1–5px:

   ```
   fraction  fine_pts  coarse_pts   max_deviation_pt
   0.02        18          6          0.0153
   0.10        92         26          0.0464
   0.25       122         32          0.0076
   0.50       164         44          0.0007
   0.75       204         54          0.0007
   0.95       280         72          0.0305
   1.00       322         82          0.0007

   Worst across all fractions and thicknesses: 0.0990 pt
   ```

   A Retina pixel is 0.5pt, so worst case is ~1/5 of a pixel.

**Result: 13–18% → 0.0–0.2%**, held over several minutes with playback confirmed active throughout. The bar still renders and tracks correctly.

One smaller thing: the overlay in `ContentView` is mounted whenever `Defaults[.showMediaProgressBar]` is true and hidden with `.opacity(isProgressBarVisible ? 1 : 0)`. Since `TimelineView(.periodic)` keeps firing regardless of opacity, it wakes twice a second even with the notch open or music paused. I passed `isProgressBarVisible` into the view and only construct the `TimelineView` when it's true, falling back to a static render otherwise — that keeps the fade transition intact while letting the timeline go away when hidden.

Happy to open a PR against your branch with these three changes if useful, or you may prefer to fold them in yourself. Either way — nice work, I'm keeping this running locally.

dstn-idly:
Small follow-up — since the rendering optimization steps the bar every 0.5s, it can look a tiny bit steppy if you look closely. So I added a "Progress bar smoothness" picker in Settings → Media (Power saver / Balanced / Smooth) that just controls how often it updates: smaller = smoother, sub-pixel steps.

It defaults to **Power saver (0.5s)** so CPU stays right where @ericrcwu001's fix put it, and the smoother options are opt-in. And since the contour sampling is already coarse now, even "Smooth" (0.1s) stays well below the old cost. 🙂

dstn-idly:
Whoa, thank you for this @ericrcwu001 genuinely one of the best reviews I've gotten, you basically did the profiling and the geometry proof for me lol.

You're totally right, I didn't clock that it was rebuilding the whole ribbon every single frame just to move it a fraction of a pixel. I went ahead and applied all three of your changes:

Dropped animatableData + the .animation on fraction
Coarsened the sampling to 12/16/12
Only running the TimelineView while the bar's actually visible

Just pushed them to the branch (commit eba55b5) and credited you in the commit. Builds clean and looks identical to me. Thanks again for taking the time to dig into this

PR #1363, opened 2026-07-01
Add charging wattage to the battery popover

isaacau502:
### What

Adds option to show the connected power adapter's rated wattage to the battery popover, inline on the status line — e.g. **"Charging: 140W"** / **"Plugged In: 140W"**.
This makes it easy to tell at a glance whether a weak brick or cable is limiting charging (e.g. grabbing a 30W charger instead of the 140W one).

Screenshot 2026-06-30 at 11 52 05 PM

Screenshot 2026-06-30 at 11 50 23 PM

### How

- Reads the adapter's rated watts via the public `IOPSCopyExternalPowerAdapterDetails()`
  (`kIOPSPowerAdapterWattsKey`), threaded through the existing  `BatteryInfo` / `BatteryEvent` flow like the other battery fields.
- The watts are appended as a separate `Text`, so the localized "Charging" / "Plugged In" strings stay intact for Crowdin.
- New **"Show charging wattage"** toggle in Battery settings (default on), matching the existing per-element battery display toggles.
- Shows only when plugged in and a wattage is reported; hidden on battery.

### Testing

Verified the popover shows the correct wattage on multiple 140W/65W/20W adapters, updates live on plug/unplug, and
hides on battery. The settings toggle shows/hides the row.

### DIFF
--- a/boringNotch/ContentView.swift
+++ b/boringNotch/ContentView.swift
@@ -315,6 +315,7 @@ struct ContentView: View {
                                     isInLowPowerMode: batteryModel.isInLowPowerMode,
                                     isPluggedIn: batteryModel.isPluggedIn,
                                     levelBattery: batteryModel.levelBattery,
+                                    maxAdapterWatts: batteryModel.maxAdapterWatts,
                                     isForNotification: true
                                 )
                             }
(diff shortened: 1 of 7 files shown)


KohliNaman:
would love to see this, great improvement!

Alexander5015 (approved):

PR #1362, opened 2026-06-30
Add Kaset as a media source

WGlegola:
### What
Adds **Kaset** ([sozercan/kaset](https://github.com/sozercan/kaset)) — a native macOS YouTube Music client — as a selectable music source. boring.notch can now display and control Kaset playback (play/pause, next/previous, seek, volume, shuffle, repeat) and toggle the like/heart, the same way it integrates Apple Music and Spotify.

### Why
Closes #1167. Kaset users had no way to surface or control playback from the notch. Kaset exposes a full AppleScript suite plus a `com.sertacozercan.Kaset.playerInfo` distributed notification, so it can be supported cleanly with a dedicated controller.

### How
- New `KasetController` (modeled on `AppleMusicController`/`SpotifyController`), reading state from Kaset's `get player info` JSON and driving transport/like via its AppleScript commands. State refreshes on a short poll plus Kaset's distributed notification.
- Registered the source in `MediaControllerType`, `MusicManager.createController`, the onboarding + Settings pickers, and added the `com.sertacozercan.Kaset` apple-events entitlement exception.
- All `@Published` state mutation is `@MainActor`-isolated, emissions are change-gated via `PlaybackState: Equatable`, and the displayed track is cleared when Kaset isn't running.

### Testing
- ✅ Builds clean (`xcodebuild -scheme boringNotch -configuration Debug build` → **BUILD SUCCEEDED**).
- ✅ Every command, JSON key, and string token verified against Kaset's `Kaset.sdef` and `ScriptCommands.swift` (transport + `seek`; `position`/`duration`/`volume` numeric types; `repeating`/`likeStatus` token casing; bundle id `com.sertacozercan.Kaset`).
- Manually exercised with Kaset running: metadata + artwork, play/pause/next/previous, seek, volume, shuffle/repeat, like toggle, and clear-on-quit.

Issue #1289, opened 2026-05-29
[FEATURE] Playback Control Source Switching and Auto stop


Jehne07:
### Is your feature request related to a problem?

When background music is playing and a new video/audio starts (e.g. YouTube), the new playback takes over media controls, making it impossible to control the previous playback.

### Describe the solution you'd like

Proposed solutions:

-  Add a way to switch the controlled playback source on the fly.
-  Add an optional feature to automatically pause the previous playback when a new one starts, then restore it once the new playback ends, with an on-the-fly toggle to enable or disable this behavior.

Ideally, both solutions could work together.

### Use cases & user scenarios

Usage examples:

- Music is playing in Spotify. A YouTube video starts playing and temporarily takes over media controls. After the video ends, Spotify playback controls are automatically restored.

-  Music is playing in the background while a short clip is opened in a browser. The user switches the active playback source on the fly without stopping either playback.

- A podcast is playing when a voice message is opened. The podcast is automatically paused and resumed after the voice message finishes.

- When multiple media sources are active at the same time

### Additional context & screenshots

<img width="615" height="193" alt="Image" src="https://github.com/user-attachments/assets/ccef93e0-1f9f-40ee-9b53-cc702dc16253" />

### Willingness to contribute

- [ ] Yes, I can write the code for this feature
- [ ] Yes, I can help with design/mockups

### Pre-submission checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have provided sufficient detail for the team to understand the request

Issue #713, opened 2025-08-26
New Music Source Option


TheComuterGuy2:
There should be a Music Source option where you can choose multiple Sources, for example, Spotify and Apple Music or YouTube Music, Spotify and Apple etc.  Something like that.

I know there is a new "Now Playing" option, but I don't like seeing myself watching a YouTube Video or a TikTok video.

Please consider adding this as a feature. TYSM for your beautiful free app

TheComuterGuy2:
Thx!

lewislarsen:
To add to this, how would @Alexander5015 feel about a whitelist or blacklist for "Now Playing" sources? You could exclude "Google Chrome" for example, to stop the app from picking up that source, and it would fallback to the next available current audio source (if available)

PR #1366, opened 2026-07-01
Non-exclusive media sources with per-connector channel gating

WGlegola:
## Summary

Two related changes to how the notch surfaces media playback:

1. **Opt-in "Now Playing" fallback** — when your chosen app-specific source (Spotify / Apple Music / YouTube Music) isn't the audible app, the notch transparently uses the generic Now Playing source instead of going idle.
2. **Per-connector channel gating** — each media source declares, per control channel, whether it's reliably supported, so controls that don't work for a given source are greyed or hidden instead of silently doing nothing (or worse, flipping their own icon while nothing happens).

Both are safe by default: the fallback is a Settings opt-in (off by default); the gating replaces the old `supportsFavorite` / `supportsVolumeControl` flags with one uniform mechanism and changes nothing for sources that already worked.

## Why

- Picking an app-specific source and then not running/playing it left the notch idle even while music played elsewhere.
- Controls like shuffle/repeat were shown even for sources that can't drive them via the generic MediaRemote pipeline — the button changed state while the app ignored the command.

### 1. Now Playing fallback (opt-in)
- New `FallbackMediaController` composes the chosen controller with a `NowPlayingController` and forwards whichever is actually playing (precedence prefers the genuinely-playing source; app launch/quit is observed via `NSWorkspace` with a short debounce to avoid thrash).
- New Settings toggle **"Use Now Playing when the selected app isn't playing"** (`Defaults.fallbackToNowPlayingWhenInactive`, default off), in Settings → Media and onboarding.

### 2. Per-connector channel gating
- New `MediaChannelPolicy` value type: for each of 8 channels (play/pause, previous, next, seek, shuffle, repeat, favorite, volume) a `ChannelSupport` of `supported` / `disabled` (shown greyed, inert) / `hidden`.
- Each controller advertises a policy; `NowPlayingController` computes it dynamically per the underlying app's bundle id (what the generic MediaRemote pipeline can reliably drive).
- `MusicManager` publishes the active policy; `NotchHomeView` gates every control off it — hidden slots collapse to a placeholder (stable toolbar geometry), disabled controls grey out, and the progress bar keeps showing position while blocking scrubbing when seek is unsupported.
- Removes the ad-hoc `supportsFavorite` / `supportsVolumeControl` flags (subsumed).

Resulting per-source behavior:

| Source | play/pause · next · prev | seek | shuffle | repeat | favorite | volume |
|---|---|---|---|---|---|---|
| Apple Music | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| Spotify | ✅ | ✅ | ✅ | ✅ | greyed | ✅ |
| YouTube Music | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| Now Playing · Apple Music | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| Now Playing · Spotify | ✅ | greyed | hidden | hidden | greyed | ✅ |
| Now Playing · other app | ✅ | greyed | hidden | hidden | greyed | greyed |

## Testing
- Builds clean (Debug, macOS, arm64).
- Manual: verified per-source control availability and the fallback flip (play in a non-chosen app → notch follows; relaunch the chosen app → it switches back; toggle off → original idle behavior).

## Screenshots / recording
image

## Related
- Relates to #713 (non-exclusive / multiple music sources)
- Relates to #1289 (switching the controlled playback source on the fly)
- Kaset media source is submitted separately in #1362 (intentionally not in this PR)


Alexander5015:
Why does it say that seek, shuffle, and repeat disabled in some way for Now Playing (besides Apple Music)? I am also confused why some are hidden and some are greyed out. 

Also, I think it would be better to make the fallback setting more generalized rather than implementing it the way you have it now. It seems like some people want this setting mainly to avoid Now Playing, while still supporting multiple music apps. A generalized setting would let people choose which apps they want to use and define their priority order.

Separately, I think people were asking for a whitelist/blacklist specifically for Now Playing, but that can be handled independently from this PR.

WGlegola:
# Media Sources: priority list + consistent control gating

v2

### 1. A priority list of media sources (replaces the single-source picker + fallback toggle)
`Also, I think it would be better to make the fallback setting more generalized rather than implementing it the way you have it now. It seems like some people want this setting mainly to avoid Now Playing, while still supporting multiple music apps. A generalized setting would let people choose which apps they want to use and define their priority order.`

This was great idea.

I reworked two things: **which** app's playback the notch shows, and **how**
it displays controls that a given app doesn't support. I generalized the old
"pick one app + a Now Playing fallback toggle" into an ordered list of sources
one fully control.

Before, you picked **one** music source (Apple Music, Spotify, YouTube Music, or
Now Playing) and could optionally flip a single switch: *"use Now Playing when
the selected app isn't playing."*

Now there's an ordered list in **Settings → Media** where you can:

- Enable **any combination** of sources (not just one).
- **Drag** them to set their priority.
- **Toggle each one independently.**

The notch shows whichever **enabled** source is currently playing, preferring the
one higher in the list. "Now Playing" is pinned to the bottom as the catch-all,
since it works with almost any app. When nothing is actively playing, the notch
keeps showing the last thing that was.

In short: instead of "one chosen app, with an optional Now Playing fallback," you
now get "as many sources as you want, in the order you want."

### 2. Consistent, greyed-out controls (no more disappearing buttons; Now Playing un-gated)

`Why does it say that seek, shuffle, and repeat disabled in some way for Now Playing (besides Apple Music)? I am also confused why some are hidden and some are greyed out.`
I went to deep the rabbit hole. The assumption is to better hide something if there is no guarantee it will work. I wont get into details, but once i thought about it with fresh eyes i've definetly overdone it :).

Now:

- A control an app doesn't support is consistently **greyed out** (visible but non-interactive) — never removed. The layout stays stable.
- **Now Playing** is treated as supporting all controls, so it no longer greys out buttons it can actually drive.

## How it works under the hood

- **`MediaChannelPolicy`** — each source declares, per control channel (play/pause, previous, next, seek, shuffle, repeat, favorite, volume), whether it's `supported` or `disabled` (greyed). Controllers now expose a single
  `channelPolicy` instead of scattered `supportsX` booleans.
- **`PriorityMediaController`** — wraps the enabled sources and arbitrates to whichever is audibly playing: first one that's *playing*, else first one that's *active*, else it freezes on the last thing shown. It forwards controls to the live source and reflects that source's channel policy.
- **`MusicManager`** builds this composite from the saved list. A one-time migration converts existing users' old settings (single source + fallback toggle) into the new list, so no one loses their configuration.
- The onboarding "choose a source" step writes into the same list.

## Compatibility

- Existing preferences migrate automatically on first launch — nothing to do.
- If Now Playing is unavailable on a given macOS version, it's dropped from the list and the app falls back to a usable source, so the player is never dead.

## Scope

- This PR is media-source selection + control gating only.
- The **Kaset** media source is a separate PR (#1362); there is no Kaset code here. Its present on screen because i'm running by own build now.

PR #1390, opened 2026-07-15
Fix volume control for external DACs with HUD replacement

### LINKED ISSUE #961: [Bug] Volume stuck between 92 and 100 when connected to an external DAC.

This app works great when its just my macbook pro on its own, but when i plug into my ProFX console that serves as the output device for my audio, for some reason this app starts limiting the volume between 92 and 100. 

Regardless of what the level was when I switch to the ProFX, as soon as I have switched and start pressing the volume keys, the level starts trending towards 92, regardless of whether I press volume up or down. 

I expected the volume to change normally.

### Boring Notch Version

v2.7.3

### macOS Version

26.2

Spotify

_No response_

_No response_

### Relevant log output

```shell

```

### Checks

louiscrc:
## Summary
- Prefer CoreAudio `VirtualMainVolume` / `VirtualMainMute` (system HUD path) instead of averaging `VolumeScalar` across Main + channels 1–4, which stuck volume near ~92–100% on many USB DACs (iFi Go Bar, FiiO, ProFX).
- Rebind volume/mute listeners when the default output device changes.
- If the current output exposes no controllable volume, pass volume keys through to macOS instead of swallowing them.

Fixes #961

## Test plan
- [x] Build Debug on `dev` with Xcode 26
- [x] With HUD replacement enabled and an external USB DAC (iFi Go Bar) as default output, volume up/down/mute change level normally
- [x] Same check on built-in speakers (regression)
- [x] Plug/unplug DAC while app is running and confirm volume still responds after device switch
- [ ] With a device that has no software volume, keys fall through to the system

### DIFF
--- a/boringNotch/components/OSD/Managers/XPC/VolumeManager.swift
+++ b/boringNotch/components/OSD/Managers/XPC/VolumeManager.swift
@@ -10,6 +10,9 @@ import Combine
 import CoreAudio
 import Foundation
 
+private let kVirtualMainVolumeSelector: AudioObjectPropertySelector = 0x766D7663 // 'vmvc'
+private let kVirtualMainMuteSelector: AudioObjectPropertySelector = 0x766D6D63 // 'vmmc'
+
 final class VolumeManager: NSObject, ObservableObject {
     static let shared = VolumeManager()
 
@@ -25,6 +28,13 @@ final class VolumeManager: NSObject, ObservableObject {
     private var previousVolumeBeforeMute: Float32 = 0.2
     private var softwareMuted: Bool = false
 
+    private var listenedDeviceID: AudioObjectID = kAudioObjectUnknown
+    private var volumeListenerAddresses: [AudioObjectPropertyAddress] = []
+    private var muteListenerAddress: AudioObjectPropertyAddress?
+    private var volumeListenerBlock: AudioObjectPropertyListenerBlock?
+    private var muteListenerBlock: AudioObjectPropertyListenerBlock?
+    private var didRegisterDefaultDeviceListener = false
+
     private override init() {
         super.init()
         setupAudioListener()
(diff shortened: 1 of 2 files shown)


boringNotch/components/OSD/Managers/XPC/VolumeManager.swift
```diff
@@ -33,6 +43,14 @@ final class VolumeManager: NSObject, ObservableObject {
 
     var shouldShowOverlay: Bool { Date().timeIntervalSince(lastChangeAt) < visibleDuration }
 
+    var canControlVolume: Bool {
+        let deviceID = systemOutputDeviceID()
+        guard deviceID != kAudioObjectUnknown else { return false }
+        return hasVirtualMainVolume(deviceID: deviceID)
+            || readValidatedScalar(deviceID: deviceID, element: kAudioObjectPropertyElementMain) != nil
+            || readValidatedScalar(deviceID: deviceID, element: 1) != nil
```
Alexander5015, line 51: Could we check whether a volume property is actually writable here, rather than using existence/read to check if we can control it? I think you can use `AudioObjectIsPropertySettable` for the same addresses used by `writeVolumeInternal`.

boringNotch/components/OSD/Managers/XPC/VolumeManager.swift
```diff
@@ -124,146 +142,219 @@ final class VolumeManager: NSObject, ObservableObject {
...
-            if let v = readValidatedScalar(deviceID: deviceID, element: el) { collected.append(v) }
+
+        if let v = readVirtualMainVolume(deviceID: deviceID) {
+            return v
+        }
+        if let v = readValidatedScalar(deviceID: deviceID, element: kAudioObjectPropertyElementMain) {
+            return v
+        }
+        var channels: [Float32] = []
+        for el in [UInt32(1), UInt32(2)] {
```
Alexander5015, line 299: Don't hardcode the channels here. You should fetch the specific channels for the device, since it is not guaranteed to be channels 1 and 2, particularly if it is not a two channel device. I think this is how you do that: https://developer.apple.com/documentation/coreaudiotypes/audiochannellayout

Alexander5015 (changes requested):

Alexander5015:
Thanks for working on this. I left some comments to improve the reliability of it.

PR #1396, opened 2026-07-19
Add optional YouTube Music (Browser) source via a first-party browser extension bridge

### LINKED ISSUE #1323: [FEATURE] Add optional Chrome extension bridge for YouTube Music PWA support
## Is your feature request related to a problem?

Users who already listen through the official YouTube Music PWA in Chrome currently need Pear Desktop or another YouTube Music desktop wrapper for BoringNotch's YouTube Music source. I would like to avoid requiring a separate third-party music app when the official PWA is already open and playing.

This can also be confusing because the source is labeled "YouTube Music", but in practice it depends on a separate YouTube Music/Pear Desktop app rather than the official Chrome PWA.

Related issues I found while checking for duplicates:

- #1252
- #1269
- #1041
- #1042
- #1032
- #763
- #1015

I did not find an exact duplicate for a first-party YouTube Music PWA Chrome extension bridge.

## Describe the solution you'd like

Add an optional first-party BoringNotch Chrome extension bridge for `https://music.youtube.com`.

The extension would observe playback state in the YouTube Music PWA, send normalized now-playing updates to BoringNotch over a local first-party bridge, and receive commands from BoringNotch for play/pause, next, previous, and later seek/volume/like.

Target architecture:

```text
YouTube Music PWA
  -> Chrome extension content script
  -> Chrome extension background service worker
  -> local BoringNotch bridge
  -> BoringNotch music controller
```

This should coexist with the current Pear Desktop integration. Existing Pear Desktop users should not lose support.

Important implementation note: Chrome extensions can open outbound WebSocket/fetch connections, but they cannot directly host a normal localhost server for BoringNotch to call. So the PWA bridge likely needs either:

- a BoringNotch-hosted local bridge that the extension connects to, or
- a bundled Native Messaging host registered by BoringNotch.

The first prototype should prove the extension can reliably read YouTube Music PWA state and receive commands before a PR is opened.

## Use cases & user scenarios

- A user installs YouTube Music as a Chrome PWA and wants BoringNotch play/pause/next/previous controls without installing Pear Desktop.
- A user wants the official YouTube Music web experience but still wants BoringNotch's notch controls and artwork.
- A privacy-conscious user prefers a small open-source bridge with narrow permissions over a separate desktop wrapper.
- A user wants YouTube Music control to recover when Chrome/PWA reloads or BoringNotch restarts.

## Additional context & screenshots

Current BoringNotch code appears to support YouTube Music through a local Pear Desktop API at `http://localhost:26538`, with HTTP commands and a WebSocket feed.

An initial scope could be:

- keep the existing Pear Desktop source intact
- add a separate optional "YouTube Music PWA" source
- add a narrow Manifest V3 Chrome extension for `https://music.youtube.com/*`
- send title, artist, artwork, duration, position, and play/pause state to BoringNotch
- support play/pause, next, and previous commands first
- add a mock page/test harness before testing the real YouTube Music PWA
- only prepare a PR after the bridge has been tested against both the mock page and the real PWA

I am willing to help write the code for this feature if the approach sounds acceptable.

## Willingness to contribute

- [x] Yes, I can write the code for this feature.
- [x] Yes, I can help with design/mockups.

## Pre-submission checklist

- [x] I have searched existing issues to ensure this is not a duplicate.
- [x] I have provided sufficient detail for the team to understand the request.

munimx:
closes #1323

## What this adds

An **optional** media source, **YouTube Music (Browser)**, that reads and controls the official YouTube Music web player at `music.youtube.com` through a small first-party browser extension.

Today the "YouTube Music" source talks to [Pear Desktop](https://github.com/pear-devs/pear-desktop) over `http://localhost:26538`. That works well, but it means anyone already listening in the YouTube Music PWA has to install a second music app to get notch controls — and the source being labelled simply "YouTube Music" makes that surprising.

**The existing Pear Desktop source is completely untouched.** This is added alongside it; users pick one in Settings → Media → Music Source.

The notch showing a track from the YouTube Music web player, with the Media settings pane reporting the browser extension as connected

*Playing in the YouTube Music web player, driven from the notch. The settings pane shows the extension connected with nothing configured — the identifier below it is read from the WebSocket handshake, not entered by the user.*

## How it works

```
YouTube Music tab / PWA window
  ├── src/player.js     MAIN world      reads #movie_player + navigator.mediaSession
  ├── src/content.js    isolated        relays page <-> service worker
  └── src/background.js service worker  WebSocket client + 20s keepalive
        │
        │  ws://127.0.0.1:26539   loopback only, Origin authorised
        ▼
  BrowserBridgeServer  →  YouTubeMusicPWAController  →  MusicManager
```

A browser extension can't listen on a socket, so the app is the server and the extension is the client. The issue floated Native Messaging as the alternative; I went with a loopback WebSocket because it needs no helper binary, no per-browser manifest installation, works across Chromium browsers unchanged, and is directly testable from a script. Native Messaging remains a reasonable future addition.

Three things worth knowing about the implementation:

- **No third-party dependency.** `NWListener` provides the socket; the WebSocket layer (handshake + framing) is ~250 lines in `BrowserBridgeWebSocket.swift`. That seemed worth keeping in-tree given the project pins its SPM packages at exact versions.
- **No entitlement changes.** `com.apple.security.network.server` is already present, and I confirmed at runtime that the sandboxed release build binds `127.0.0.1` with it.
- **The server is owned by the controller**, so the listener only exists while this source is selected. No `AppDelegate` changes, and no socket open for users who never enable it.

## Security, and why there's nothing to pair

The listener binds `127.0.0.1` only — but loopback is reachable by every local process, including any web page the user has open, so that alone is not access control.

The check that does the work is the **`Origin` header of the WebSocket handshake**. The browser sets it and page script cannot override it:

| Caller | `Origin` | Result |
|---|---|---|
| the extension | `chrome-extension://<id>` | accepted |
| any website | `https://…` | **403 at the handshake** |
| `curl` / a script | none | **403 at the handshake** |

This is the reason the setup is zero-touch: the check that keeps websites out needs nothing from the user — no token, no prompt, no port. It is also why the app speaks RFC 6455 itself. `NWProtocolWebSocket` would have been less code, but its server-side metadata exposes only opcode/response/closeCode and gives no access to the handshake headers, which would have forced a user-entered secret instead.

Also:

- A connection that fails the origin check **never becomes the active client**, so it cannot disconnect a working extension. I hit the inverse of this in an earlier revision — any local process could kick the extension off by repeatedly connecting — so unauthorised connections now sit in a bounded pending list.
- A native process could forge `Origin`, but it could already synthesise media keys, so this grants it nothing new.

Extension permissions are `storage`, `alarms`, and `https://music.youtube.com/*`. No `tabs`, no `<all_urls>`.

## Setup and distribution

**For users:** install the extension, pick the source in Settings, open YouTube Music. That's the whole flow. If the default port is busy the extension tries 26540 and 26541 and remembers what worked, so a port clash resolves itself rather than becoming a support question.

**For maintainers:** `chrome-extension/package.sh` produces a Web Store zip. There's no build step and no dependencies, so the reviewed source and the shipped source are byte-identical. The app doesn't need to know the published extension ID — authorisation is by origin *scheme*, so a Web Store build, a locally loaded build, and a build running in Brave or Edge all work with no configuration change.

## Reading YouTube Music's state

Verified against the live site. Several of the obvious sources turn out to be unreliable, which is worth recording:

| What | Source used | Why not the obvious one |
|---|---|---|
| playing / paused | `navigator.mediaSession.playbackState` | `movie_player.getPlayerState()` reports buffering/unstarted during normal playback, and `#play-pause-button`'s `title` goes stale |
| title / artist / album | `navigator.mediaSession.metadata` | the DOM byline reads "Artist • 613M views • 3.4M likes" on video entries |
| duration | `movie_player.getDuration()` | `video.duration` is `NaN` briefly after every track change |
| volume | `movie_player.getVolume()` (0–100) | `video.volume` is 0–1 |
| repeat | `ytmusic-player-bar[repeat-mode]` | — |
| favourite | `ytmusic-like-button-renderer[like-status]` | — |

`player.js` runs in the MAIN world because `#movie_player` is a page global. All selectors live in one `SEL` object so a YouTube Music redesign is a single-file fix.

## Testing

Built with Xcode 26.6, **Release** configuration (what CI uses) — clean, no new warnings.

Verified end-to-end with the real extension, the real site, and the running app:

- **Read** — notch shows live title, artist, artwork and position; state streams at 1 Hz.
- **Write** — every command driven into the page and confirmed by reading page state independently: `play`, `pause`, `toggle` (both directions), `next`, `previous`, `seek` (landed at 120.8s), `setVolume` (`getVolume()` returned exactly 35 for 0.35), `toggleRepeat` (NONE → ALL), `toggleShuffle`.
- **Resilience** — killed the app mid-session and the extension reconnected on its own; restarted again, reconnected again; survived 50s fully idle, confirming the 20s keepalive beats the 30s MV3 service-worker timeout.
- **Regression** — the Pear Desktop source is not modified by this change.

Because the WebSocket layer is hand-written, it has its own raw-socket conformance suite (15 checks, all passing):

- websites and origin-less clients refused with 403; `chrome-extension://` and `moz-extension://` accepted
- `Sec-WebSocket-Accept` verified against an independent SHA-1 implementation
- fragmented messages and 16-bit-length payloads reassembled without desync — asserted by requiring a subsequent ping to still be answered, which only holds if the parser consumed exactly the right byte count
- two frames coalesced into one TCP segment both parsed; one frame split across segments buffered
- unmasked client frames (a protocol violation) drop the connection
- server frames confirmed unmasked

`setFavorite` is implemented and its transport is verified, but I did not fire it against a real account since it would like a song in the user's library.

For reviewers who'd rather not install anything, `chrome-extension/test/bridge-sim.mjs` impersonates the extension over the same protocol (Node 22+, no dependencies). Note that Node's WebSocket client sends no `Origin`, so a stock build refuses it by design — allow a test origin in `BrowserBridgeHandshake.isAllowedOrigin` to use it.

## Notes for reviewers

- **No translation changes.** New strings are plain `Text(...)` / `NSLocalizedString` for automatic extraction; `Localizable.xcstrings` is untouched.
- The extension lives in `chrome-extension/` so the wire protocol and both of its implementations stay versioned together. Happy to split it into its own repository if you'd prefer not to carry JavaScript here.
- Naming: I used "YouTube Music (Browser)" to keep it distinguishable from the existing entry without renaming that one. Open to something clearer.
- The hand-written WebSocket layer is the one part I'd most want a careful look at. It exists specifically to read `Origin`; if you'd rather keep `NWProtocolWebSocket`, the fallback is a pairing token or a first-run approval prompt, both of which put the setup cost back on the user.
- I'm aware #1323 hasn't been triaged yet, so treat this as a concrete proposal to react to rather than something expecting a merge as-is. Happy to change the approach, the scope, or drop it if it's not a direction you want.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jooapa:
W song

Issue #1312, opened 2026-06-10
[BUG] Drag and drop


Shivs-07:
### Is your feature request related to a problem?

this feature is working properly until I try on WhatsApp. 
when I drag and drop file to WhatsApp it gives me the address(file location) of the file in my Mac. 
till now I have noticed this problem this issue in this WhatsApp only.
 

### Describe the solution you'd like

So, what I want is like drag and drop should be universal , it should work. in all the app and web pages. 
just fix this issue and check whether this issue is found on other apps also.

### Additional context & screenshots

https://github.com/user-attachments/assets/5956750f-1c68-4ec3-a64d-998874c2edd8

Issue #1133, opened 2026-04-06
[Bug] Can't Drag and Drop into Affinity


sherpascott:
### What happened? What are the steps to reproduce the issue?

I can drag and drop png files from Finder into Affinity, but when dragging a png from the notch, it does not input the image and instead just pastes the file location. 

### What did you expect to happen?

I expected the png file to input as an image. I would love to have this fixed as I prefer to use Affinity in fullscreen mode while using the notch to import files.

### Boring Notch Version

v2.7.3

### macOS Version

Tahoe 26.4

### Music Source? (If relevant)

None

### Relevant log output

```shell

```

### Checks

- [x] I haven't found any duplicates with my issue.

PR #1398, opened 2026-07-21
Fix shelf file drag-and-drop into WhatsApp

ShehanCes:
Dragging a file out of the shelf into WhatsApp inserted the file's path as text instead of attaching the file.

The drag payload wrote the file URL as a string alongside a plain-text path, so apps like WhatsApp picked up the path text. It now writes the URL itself, matching the shelf's Copy behavior, so the actual file is attached.

Fixes https://github.com/TheBoredTeam/boring.notch/issues/1312

https://github.com/user-attachments/assets/b0b7687f-482c-46bb-a09c-de3db5434d81


Ein-Tim:
- Does this also fix #1133?

ShehanCes:
not sure if it fixes affinity, someone will have to check and see.

Ein-Tim:
Ok thanks, once the change is released I will follow up in the issue.

Issue #1400, opened 2026-07-22
Discussion: current lyric in the closed notch + word-level (Enhanced-LRC / YRC) karaoke fill


rodrigofdzr:
**Context**

Per CONTRIBUTING ("discuss major changes first"), I'd like to float a couple of lyric enhancements I've prototyped in a fork before opening any PR — and to coordinate with the lyric work already in flight (#937, #992) so we don't collide.

As I understand current `dev`: `LyricsService` provides line-level synced lyrics (LRCLIB + Apple Music AppleScript), rendered in the **open** notch via `TimedLyricText` (a nice single-glide reveal-tail scroller), gated by `enableLyrics`. There's no lyric text in the closed notch, and no word-level timing.

**Proposal**

1. **Current lyric line in the _closed_ notch.** Show the active line in the always-visible closed strip, reusing the existing `TimedLyricText` scroller, behind an opt-in setting (default off).
   - Design constraint I want to confirm first: on notched Macs the closed strip's centre is the camera housing, so there's no room for lyric text without widening the closed notch (a persistent version of the inline sneak-peek expansion). On notchless displays the space is already there. **Is a persistent closed-notch expansion for lyrics something you'd accept, or would you prefer this gated to notchless displays only / not at all?** I'd rather agree on the approach before building, since it touches the always-visible layout.

2. **Word-level karaoke fill.** A gray→white progressive fill on the active line, driven by inline `<mm:ss.xx>` word timings (Enhanced-LRC / Netease YRC), with a linear per-line fill fallback when only line-level timing is available.

**Coordination with in-flight PRs**

- **#937 (Netease + improved LRC parsing):** a word-level fill needs a source that carries word timings — Netease's YRC is the natural provider, so this builds directly on #937's direction rather than duplicating it. Does #937 intend to expose word-level timings, or line-level only?
- **#992 (lyrics shimmer + closed-notch visualizer):** the karaoke fill and the shimmer/shine are competing visual treatments on the active line, and #992 also touches the closed-notch right-side indicator. Should these be unified (shimmer as the chosen "active line" effect), or is there room for a distinct karaoke-fill option?

I have working prototypes of both (single-glide closed-notch scroll + character-weighted word fill) and am happy to reimplement cleanly against the current `LyricsService` if there's interest. Mainly want to align on direction first. Thanks!

Alexander5015:
[...]

PR #1402, opened 2026-07-23
feat: show current lyric below the closed notch (independent toggle)

rodrigofdzr:
## Summary

Adds an opt-in setting to show the **current lyric line while the notch is closed**, independent of the existing open-notch lyrics. Discussed in #1400; design reworked per maintainer feedback in this PR (sneak-peek style, no separate capsule).

## Behavior

- **Notched displays:** the line renders sneak-peek style in the closed strip. A single continuous line would hide its middle under the camera, so it's **split at the word boundary that best balances both halves' rendered widths**, flanking the camera (like the inline sneak peek's title/artist). Symmetric flanks keep the gap centered under the camera, and the **strip grows/shrinks with each line** — full font size, no word ever cut.
- **Non-notch displays:** the line renders **inline in the live-activity strip** with the existing `TimedLyricText` reveal-tail scroll.
- **Independent toggle** `showLyricsInClosedNotch` (default **off**) in Media settings; the lyrics fetch triggers when either lyrics toggle is on, and toggling takes effect immediately for the current track (also fixes that pre-existing papercut for the open-notch toggle).
- A track-change sneak peek takes priority over the lyric; zero footprint when inactive.

## Recordings

See [this comment](https://github.com/TheBoredTeam/boring.notch/pull/1402#issuecomment-5144892016) for screen recordings on both a notched MacBook and a notch-less display, plus the settings UI.

## Testing

Built and verified on both display types: line-by-line strip resizing, seeks/track changes, pause/stop, toggle on mid-song takes effect immediately, existing open-notch lyrics and sneak peeks unchanged.

### DIFF
--- a/boringNotch/ContentView.swift
+++ b/boringNotch/ContentView.swift
@@ -431,6 +431,150 @@ struct ContentView: View {
         )
     }
 
+    // Whether the current lyric should occupy the closed live-activity strip:
+    // opt-in, playing, synced lyrics available, and no inline sneak peek active
+    // (a track-change sneak peek takes priority over the lyric).
+    private var shouldShowClosedNotchLyric: Bool {
+        Defaults[.showLyricsInClosedNotch]
+            && musicManager.isPlaying
+            && !LyricsService.shared.syncedLyrics.isEmpty
+            && !(coordinator.expandingView.show && coordinator.expandingView.type == .music)
+    }
+
+    private var inlineSneakPeekActive: Bool {
+        coordinator.expandingView.show
+            && coordinator.expandingView.type == .music
+            && Defaults[.sneakPeekStyles] == .inline
+    }
+
+    private let nonNotchLyricWidth: CGFloat = 200
+
+    // Split a lyric line at the word boundary that best balances the two halves
+    // by RENDERED WIDTH (not word count), so the symmetric flanks waste as
+    // little edge space as possible. Never cuts a word.
+    private func splitLyricForNotch(_ text: String) -> (left: String, right: String) {
+        let words = text.split(separator: " ", omittingEmptySubsequences: true).map(String.init)
+        guard words.count > 1 else { return (text, "") }
+        let font = NSFont.preferredFont(forTextStyle: .caption1)
+        let attrs: [NSAttributedString.Key: Any] = [.font: font]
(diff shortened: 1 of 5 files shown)


Alexander5015:
Please add screenshots or a screen recording to all UI-related PRs, so we can see what the changes look like. 

I haven't had a chance to run it yet, but it looks like you are planning to make it a separate capsule shape below the notch? If I am correct, the idea was that the text would appear the same way default sneak peek works, sorry I should have been clear about that. It should also look different depending on if there is a notch or not, for displays without it should display as you had originally intended.

rodrigofdzr:
Thanks for the feedback — you were right, the separate capsule was a misread on my part. Reworked per your guidance, and screenshots/recordings below.

## Final design (updated in the PR)

- **Notched displays:** the current line renders **sneak-peek style** in the closed strip. Since a single continuous line would hide its middle under the camera, the line is **split at the word boundary that best balances the two halves' rendered widths**, and the halves flank the camera (like the inline sneak peek does with title/artist). The flanks are symmetric so the gap always stays centered under the camera, and the **strip grows/shrinks with each line** — full font size, no word ever cut.
- **Non-notch displays:** the line renders **inline in the live-activity strip** (between the album art and the visualizer) with the existing `TimedLyricText` reveal-tail scroll.
- **Independent toggle** (`showLyricsInClosedNotch`, default off) in Media settings, separate from the open-notch lyrics — and toggling either lyrics setting now takes effect immediately for the current track.
- A track-change sneak peek still takes priority over the lyric.

Tested on both a notched MacBook and a notch-less display. Branch squashed to a single clean commit.

## Recordings

- ▶️ [Notched display — split lyric flanking the camera (Notch.mov)](https://github.com/rodrigofdzr/boring.notch/blob/pr-1402-media/Notch.mov)
- ▶️ [Notch-less display — inline lyric in the strip (NoNotch.mov)](https://github.com/rodrigofdzr/boring.notch/blob/pr-1402-media/NoNotch.mov)

## Settings

Settings toggle

rodrigofdzr:
https://github.com/user-attachments/assets/683fdf68-b6fb-4509-9ff1-b4b2defe1ee5

SettingNotch

https://github.com/user-attachments/assets/204ca8f7-b6ba-4817-aecb-7afd49091d7c

PR #1404, opened 2026-07-24
fix(media): resolve stale artwork in browser private browsing and helper processes(fix for macOS 26.5.2)

### LINKED ISSUE #1394: [Bug] Media information disappears when switching between multiple active players

**Description**

Boring Notch does not correctly fall back to another active media session after the most recently displayed media is paused.

**Steps to reproduce**

1. Start playing a song on Spotify.
2. Confirm that the Spotify song appears in Boring Notch.
3. While the Spotify song is still playing, start another song in Apple Music.
4. Boring Notch correctly switches to the Apple Music song.
5. Pause the Apple Music song while the original Spotify song continues playing in the background.

**Expected behavior**

After pausing Apple Music, Boring Notch should detect that Spotify is still playing and switch back to displaying the Spotify song.

**Actual behavior**

Boring Notch disappears as if no media is playing, even though the Spotify song is still playing in the background.

**Additional issue noticed**

I also noticed a related artwork issue while testing:

When I opened a YouTube video in a Safari Private Browsing window, the media appeared in Boring Notch, but the thumbnail was incorrect. Instead of showing the video thumbnail or a generic private-browsing icon, it displayed the album artwork from the previously played Apple Music track.

I also tested the same scenario in Chrome Incognito mode. In Chrome, Boring Notch displayed the Incognito icon as the media artwork, so this behavior seems to occur only with Safari Private Browsing.

My assumption is that Safari may not provide any artwork for YouTube videos opened in Private Browsing, causing Boring Notch to keep using the last available artwork instead of clearing it or displaying a fallback image. I understand that the missing thumbnail may be browser-related, but it would still be better if Boring Notch did not reuse outdated artwork in this situation.

**Video**

I recorded a video showing both issues but I wasn’t able to attach the video directly to the issue, so I uploaded it to Google Drive:
[https://drive.google.com/file/d/1UsKUyCOx-nvj6TAnKiLndo2751OnCESy/view?usp=sharing](https://drive.google.com/file/d/1UsKUyCOx-nvj6TAnKiLndo2751OnCESy/view?usp=sharing)

### Boring Notch Version

v2.7.3

### macOS Version

26.5.2

Now Playing

Spotify, Apple Music

Youtube (private browsing)

### Relevant log output

```shell

```

### Checks

SpaceCypher:
Fixes #1394

### Why We Need This
In macOS (specifically on macOS 26.5.2 Tahoe with updated WebKit process isolation), media played inside Safari Private Browsing or browser helper processes is attributed to sandboxed subprocess identifiers (such as `com.apple.WebKit.WebContent` or `.helper` processes). 

Because `com.apple.WebKit.WebContent` is a background helper framework rather than an application bundle, `AppIconAsNSImage` failed to resolve an app icon for the browser, causing stale artwork from a previously played track (e.g., Apple Music or Spotify) to remain stuck on screen.

### What Was Changed
1. **Bundle Identifier Normalization** (`AppIcons.swift`): Added a `normalizeBundleIdentifier` helper function to map WebKit subprocess identifiers (`com.apple.WebKit.WebContent`, etc.) and Chromium/Electron helper processes back to their main application bundle identifiers (e.g., `com.apple.Safari`).
2. **Default Artwork Fallback** (`MusicManager.swift`): Added a fallback in `MusicManager.swift` so that if no artwork metadata and no app icon can be resolved for a media session, the artwork is explicitly reset to the default placeholder image (`defaultImage`), ensuring stale artwork is never left stuck on screen.

---

### How Was This Tested
1. Played a track in Apple Music/Spotify until artwork appeared in the Boring Notch.
2. Opened a YouTube video in a **Safari Private Browsing** window (`com.apple.WebKit.WebContent` session).
3. Verified that the old Apple Music album cover was cleared and the Notch correctly resolved and displayed the **Safari app icon**.

### DIFF
--- a/boringNotch/helpers/AppIcons.swift
+++ b/boringNotch/helpers/AppIcons.swift
@@ -36,10 +36,29 @@ struct AppIcons {
     
 }
 
+func normalizeBundleIdentifier(_ bundleID: String) -> String {
+    let lower = bundleID.lowercased()
+    
+    // Handle WebKit / Safari rendering helper processes
+    if lower.contains("safari") || lower.contains("webkit") {
+        return "com.apple.Safari"
+    }
+    
+    // General rule for Chromium/Electron helper processes
+    // e.g., "com.google.Chrome.helper" -> "com.google.Chrome"
+    let components = bundleID.components(separatedBy: ".")
+    if let helperIndex = components.firstIndex(where: { $0.lowercased() == "helper" }) {
+        return components[0..<helperIndex].joined(separator: ".")
+    }
+    
+    return bundleID
+}
+
 func AppIcon(for bundleID: String) -> Image {
     let workspace = NSWorkspace.shared
+    let normalizedID = normalizeBundleIdentifier(bundleID)
     
-    if let appURL = workspace.urlForApplication(withBundleIdentifier: bundleID) {
+    if let appURL = workspace.urlForApplication(withBundleIdentifier: normalizedID) {
         let appIcon = workspace.icon(forFile: appURL.path)
         return Image(nsImage: appIcon)
     }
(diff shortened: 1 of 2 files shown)


Alexander5015 (approved):

SpaceCypher:
thank you.

PR #1421, opened 2026-07-30
Add real-time audio-reactive spectrum visualizer

UM7lab:
## Summary

Replaces the closed-notch waveform's random animation with bars driven by the audio that is actually playing.

- Captures system audio through a private Core Audio process tap (macOS 14.2+, `CATapDescription` + a private aggregate device), reads it via an `AudioDeviceIOProc`, and turns it into six octave-wide frequency bands with a live FFT (Accelerate/vDSP). Nothing is recorded or written to disk — samples live in a ring buffer and get thrown away after analysis.
- The loudness window is self-calibrating: it follows a slow (~6.5s) running average of the track instead of a fixed dB threshold. A fixed threshold calibrated against one track consistently looked wrong on another (mastering loudness, genre, and output device all shift the absolute levels the tap sees), so anchoring to the material itself removes that guesswork.
- An optional **Dynamic sensitivity** toggle (on by default) widens that window in real time whenever the loudest band would otherwise pin at full height, so a heavily compressed chorus keeps moving instead of flattening into a straight line for the few seconds it takes the slow average to catch up — then relaxes back once the level settles.
- `MusicVisualizer` now draws pixel-snapped `Capsule` bars instead of a `CAShapeLayer` path masked by an opaque `backgroundColor`. That combination silently discarded the rounded corners whenever the layer was scaled for animation, and non-integer bar widths made adjacent bars look like they had different thicknesses due to inconsistent anti-aliasing.
- Falls back to the original random animation whenever the tap can't be started (older macOS, permission denied, HAL error), with the reason surfaced in Settings rather than failing silently.

New Settings toggles under **Appearance → Media**: "Real-time audio spectrum" (off by default, since first use prompts for the system-audio-recording permission) and "Dynamic sensitivity" (on by default).

Requires `NSAudioCaptureUsageDescription` in Info.plist and the `com.apple.security.device.audio-input` entitlement, both added here.

## Test plan

- [x] Builds cleanly via `swift build` against this isolated diff on top of upstream `main`
- [ ] Manual test on macOS 14.2+: toggle "Real-time audio spectrum" on, grant the system-audio permission prompt, confirm bars follow music instead of animating randomly
- [ ] Confirm bars stay proportional (no clipping) through quiet passages and loud/compressed choruses with "Dynamic sensitivity" on
- [ ] Confirm the fallback (random animation) still works with the toggle off, and on macOS < 14.2
- [ ] Confirm the six bars render with even, pixel-consistent thickness at a few different notch heights

### DIFF
--- a/boringNotch.xcodeproj/project.pbxproj
+++ b/boringNotch.xcodeproj/project.pbxproj
@@ -80,6 +80,7 @@
 		1443E7F32C609DCE0027C1FC /* matters.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1443E7F22C609DCE0027C1FC /* matters.swift */; };
 		147163982C5D35B70068B555 /* MusicVisualizer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 147163972C5D35B70068B555 /* MusicVisualizer.swift */; };
 		1471639A2C5D35FF0068B555 /* MusicManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 147163992C5D35FF0068B555 /* MusicManager.swift */; };
+		AA01A0DE2F00000000000002 /* AudioSpectrumManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA01A0DE2F00000000000001 /* AudioSpectrumManager.swift */; };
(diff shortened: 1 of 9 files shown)


UM7lab:
Closing — while working on this I discovered that #1214 already merged a full real-time FFT audio visualizer into `dev` back in April, which this PR duplicates. Sorry for the noise; I based this off `main` and missed that `dev` had already moved well past it.

Issue #905, opened 2025-12-10
[FEATURE] Improve macOS Installation Guidance


Alexander5015:
**Is your feature request related to a problem? Please describe.**
For newer macOS users(or even some more experienced users), the current installation process can be confusing. When opening the disk image, there is no clear indication that the app should be dragged into the Applications folder. Additionally, the app does not prompt users on launch to move it to Applications, unlike some other macOS apps.

**Describe the solution you'd like**
Add icon for Applications folder and background image with arrow to better indicate the intallation process. Especially for newer macOS users, this is quite confusing as they run the app from the disk image. Also consider adding a dialog offering to move it as well if possible.

**Checks**
- [x] I haven't found any duplicates with my issue.

selfsta:
I would like to work on this issue.

Alexander5015:
@selfsta we already have an arrow image that will make it clearer that you should drag, we just haven't made a new release that uses it. We could use some help with a PR for the dialog though.

selfsta:
I can help. Is there an open issue or a list of changes that are needed?

Alexander5015:
This should be the only open issue for this. While some apps auto-detect if they aren't in the `/Applications` folder and offer to move themselves, that is usually for zip distributions and I just left it in as a potential option. It shouldn't be necessary for us since our release disk image has a clear background showing the /Applications alias, but if you want to try to make a PR for this, I would probably accept it, if it is pretty lightweight and simple. The creation and distribution of our releases are handled automatically by our release.yml GitHub Actions workflow if you want to see the dmg background image process.

If you want to contribute in general, and are not sure where to start feel free to ping me on Discord to get a sense of the priorities. We welcome most bug fix PRs, so opening a PR to fix an issue you are having or looking through open bug reports, especially the confirmed ones, can be a good place to start. Just note that our issue statuses usually lag behind the `dev` branch, as issues stay open until the fix is officially released (like this one). I plan to improve this process with better labels when code is merged and mentioning issues in PRs that resolve those issues at some point.

Alexander5015:
I can't remember any of those zip distributed apps that do that off of the top of my head, but maybe Crossover was one?

marcoo93:
As I mentioned in the other post about this issue, opening the dmg and dragging the app into the Applications folder creates a shortcut instead of actually installing the app.
What worked for me was: open the dmg, select the app and copy it, then open the Applications folder and paste it.
You can immediately check the installation is correct by looking at the app icon.
[...]

PR #1448, opened 2026-08-13
Prevent running from the installer and improve first-run setup

devsticks:
## Summary

- Stop Boring Notch from continuing when it is launched from a read-only installer volume or App Translocation.
- Explain why it must be copied, open Applications, highlight the mounted copy when possible, and quit before setup begins.
- Ask new users whether Boring Notch should start at login, with an explicit **Not Now** option.
- Clarify manual installation and Gatekeeper steps, keeping the one-line Terminal method recommended and System Settings as the alternative.

## Why

Even with the Applications shortcut now present on `dev`, users can still open the app directly from the mounted disk image. That copy disappears when the image is ejected and will not be available after a restart. The launch-at-login preference was also only discoverable later in Settings.

Relates to #905.

### Launched from the installer

Alert explaining that Boring Notch must be moved to Applications

### Start at Login onboarding

First-run choice to start Boring Notch at login

## Validation

- `xcodebuild` Debug build for macOS with code signing disabled
- Built a fresh test DMG using the repository's hash-pinned packaging dependencies
- Verified the DMG checksum, Applications alias, upstream background, and packaged app
- Launched the packaged app from the mounted read-only DMG and verified the location guard appears before setup
- Verified the guarded exit leaves no MediaRemote helper processes and the disk image ejects cleanly
- Walked the complete first-run onboarding with isolated preferences and verified the new startup choice and **Not Now** completion path
- `git diff --check`
- No translation files changed; new source strings can sync through Crowdin


Alexander5015:
What is the difference between the DMG backgrounds? 

Also, I think the terminal flow should be prioritized over the system settings flow as it is simpler and works reliably, unlike the system settings flow which very often does not work and leads to confusion for some people.

devsticks:
@Alexander5015 great catch; my apologies. I see the background was actually already present as of [eeef0e6](https://github.com/TheBoredTeam/boring.notch/commit/eeef0e69304ec2bc7ab29047cb36d712765dec79) (12 Dec '25), but I'd rushed to implementing a fix based on my experience installing 2.7.3, which excludes this. 

I've removed the DMG background change, and have reverted to the terminal-focused flow. 

PR out of draft now 👊

devsticks:
Also happy to remove the README.md working alterations. I realise they may muddy the review waters.

Alexander5015:
It's fine to keep it as is. I do like most of the README alterations, but I would like to tweak it. I will send a review later today with some README wording suggestions and possibly comments on the launch on login part after I review that code.

boringNotch/components/Onboarding/OnboardingView.swift
```diff
@@ -170,13 +172,27 @@ struct OnboardingView: View {
...
+            case .startupPreference:
+                PermissionRequestView(
+                    icon: Image(systemName: "arrow.clockwise.circle.fill"),
+                    title: "Start Boring Notch at Login?",
+                    description: "Boring Notch lives in your menu bar. Start it automatically when you sign in so it is ready after every restart.",
+                    privacyNote: "Recommended. You can change this at any time in Settings → General.",
+                    allowButtonTitle: "Start at Login",
+                    onAllow: {
+                        LaunchAtLogin.isEnabled = true
+                        finishOnboarding()
```
Alexander5015: **[P2] Don’t finish when registration fails**

`LaunchAtLogin.isEnabled` catches `SMAppService` registration errors internally. This code always calls `finishOnboarding()`, so a denied or failed registration clears `firstLaunch` and reports completion even though launch at login remains disabled. Check the resulting state and show an error, or call `SMAppService` directly so the failure can be handled.

boringNotch/components/Onboarding/OnboardingView.swift
```diff
@@ -170,13 +172,27 @@ struct OnboardingView: View {
...
                 )
                 .transition(.opacity)
 
+            case .startupPreference:
+                PermissionRequestView(
+                    icon: Image(systemName: "arrow.clockwise.circle.fill"),
+                    title: "Start Boring Notch at Login?",
+                    description: "Boring Notch lives in your menu bar. Start it automatically when you sign in so it is ready after every restart.",
+                    privacyNote: "Recommended. You can change this at any time in Settings → General.",
+                    allowButtonTitle: "Start at Login",
```
Alexander5015: **[P2] Route new UI copy through localization**

The title, description, and privacy note pass through `String`, so localization extraction emits only `Start at Login`; none of the new keys is present in `Localizable.xcstrings`. The installer `NSAlert` strings in `boringNotchApp.swift:503–506` are also raw strings. Non-English users therefore see this flow in English, and Crowdin cannot translate the missing keys. Use `LocalizedStringKey` or `String(localized:)` and add the source catalog entries.

boringNotch/components/Onboarding/OnboardingView.swift
```diff
@@ -170,13 +172,27 @@ struct OnboardingView: View {
...
                     }
                 )
                 .transition(.opacity)
 
+            case .startupPreference:
+                PermissionRequestView(
+                    icon: Image(systemName: "arrow.clockwise.circle.fill"),
+                    title: "Start Boring Notch at Login?",
+                    description: "Boring Notch lives in your menu bar. Start it automatically when you sign in so it is ready after every restart.",
+                    privacyNote: "Recommended. You can change this at any time in Settings → General.",
```
Alexander5015: This isn't a privacy note

PR #1455, opened 2026-08-16
Restore star history chart in README

FaintFlower:
The star history chart in the README has been missing since GitHub restricted the stargazer API and broke it (see 74ac667). This change brings the chart back with a working source. A star history chart is valuable for showcasing the project's growth, so fixing it is important.


Alexander5015:
I'm not sure why this PR is needed. Is the star chart not working right now? It should be fixed because of the change made in the commit you referenced.

Issue #1248, opened 2026-05-12
"Clear All" Keyboard shortcuts and button for File Drops


princip:
### Is your feature request related to a problem?

Currently to the best of my knowledge, managing items in the "drop files" section of the notch requires a right-click and selecting "Remove" for every item. This becomes tedious when managing multiple files or trying to quickly clear the workspace.

### Describe the solution you'd like

I would like to see more a efficient way to manage and clear the dropped files section:
1. Keyboard Shortcuts: The ability to quickly highlight an item (or multiple items - perhaps with mouse selection) and press Backspace or Delete to remove them from the notch.
2. Clear All Button: A dedicated "Clear All" button within the file drop interface to empty the notch of all items in a single click.

### Use cases & user scenarios

Users who frequently move multiple assets or documents between apps would benefit from significantly increased speed and workflow efficiency.

### Willingness to contribute

- [ ] Yes, I can write the code for this feature
- [ ] Yes, I can help with design/mockups

### Pre-submission checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have provided sufficient detail for the team to understand the request

lasersushi:
@princip I think this is being addressed in pull request #975.

princip:
@lasersushi Cheers for bringing this to my attention!

PR #975 looks like a great addition, especially the "Clear All" button and the file counter. However, it doesn't look like that PR includes the keyboard shortcuts (Backspace) for removing specific items or selections. I’d still love to see those implemented as well to make the shelf management feel even more efficient for macOS.

lasersushi:
@princip I requested that the author of the pr adds the shortcuts you requested.

Ben-Lloyd05:
@lasersushi @princip Looks like a good idea. I'm currently pretty busy but will try to work on this on the weekend / early next week. My PR is a bit out of date now and has a few merge conflicts; I'll work on getting it all up to date so it may take a little longer.

lasersushi:
@Ben-Lloyd05 Yeah, no rush. If you want help just message me.

princip:
Thanks @Ben-Lloyd05  and @lasersushi  That sounds great looking forward to testing it out once the PR is updated.

princip:
Hey @Ben-Lloyd05 have you had a chance to think about this at all?
Just wondering as I could maybe take a look at it myself, but I'm not sure I'm super well placed to do it tbh

pcruiher08:
[...]

PR #1468, opened 2026-08-17
Add shelf removal controls for the selection and the whole shelf

pcruiher08:
Relates to #1248 — delivers the Clear All half. The keyboard half is deliberately not implemented; reasoning below so nobody repeats the attempt.

## What

Removing shelf items required a right-click and "Remove" on each one, which #1248 describes as tedious when clearing several at once.

This adds one control to the shelf panel, scoped to whatever is selected:

- **With a selection** — reads `Remove Selected (n)` and removes only those items. It stays visible, because it is then the only way to act on that selection.
- **With no selection** — reads `Clear All`, revealed on hover so it does not permanently sit over the last item's tile. The panel is only as wide as the open notch.

Bulk removal is added on the view model (`remove(_: [ShelfItem])`, `removeAll()`) so the shelf is persisted once per operation rather than once per item.

Two files, +73 lines, no new dependencies, no translation changes.

## Screenshots

**No selection** — `Clear All`, revealed on hovering the panel:

Shelf panel with three items and a Clear All button in the top-right corner

**Two items selected** — the same control becomes `Remove Selected (2)` and acts only on those two:

Shelf panel with two of three items selected and the button reading Remove Selected (2)

## Why the control is scoped, rather than always "Clear All"

An earlier revision showed only `Clear All`. With multi-select already supported that turned out to be a trap: select three of eight items, and the single visible button silently removes all eight. Scoping both the label and the action to the current selection means the visible control never removes more than was just asked for.

## On the keyboard half of #1248

#1248 asks for two things: Delete/Backspace on the selection, and a Clear All button. This PR delivers the second and skips the first. The reasoning, in case it saves someone else the detour:

`BoringNotchSkyLightWindow` sets `canBecomeKey = false`, so nothing inside the notch can receive a `keyDown`. The seemingly obvious fix — take key focus while the shelf has a selection, since the window is a `.nonactivatingPanel` and can hold key status without activating the app — was implemented and tested. It is a one-way door: `resignKey()` is only a notification hook, which AppKit calls to *tell* a window it has lost key status. Calling it does not hand key status to anything else, so the panel keeps the keyboard indefinitely and the user's keystrokes stop reaching the frontmost app. Confirmed by testing, then reverted.

The mechanism that would work is a consuming `CGEvent` tap, as `MediaKeyInterceptor` already does. That requires Accessibility authorization, which seemed a steep price to pay for removing a shelf item, so I would rather you decide than spend that permission unilaterally. Happy to add it if you would prefer.

## Testing — please read, I have not run this exact branch

`dev` requires Xcode 26, which I do not have (I am on 15.2 / Xcode 16.3). I could not build or run this branch locally at all.

Rather than submit untested behaviour, I backported the identical change onto a `main`-based build, which does compile on Xcode 16.3, and tested it there:

- multi-select, then `Remove Selected (n)` removes exactly the selection and nothing else
- with no selection, `Clear All` empties the shelf
- the button is not swallowed by the panel's existing tap-to-clear-selection gesture

**The screenshots above are from that backported build, not from this branch.** The shelf code being exercised is the same — the cherry-pick onto `dev` applied cleanly — but I want to be explicit that the artefact I ran is not the artefact I am asking you to merge.

This branch itself is verified by CI only: green on all three matrix legs (Xcode `^26`/macos-26, `~26.0`/macos-15, `^27`/xcode-27). That confirms it compiles on your toolchain; it does not confirm behaviour.

If you would rather have behaviour confirmation on `dev` proper before considering this, say so and I will get onto a machine that can build it.

## Notes

Touches only `ShelfView.swift` and `ShelfStateViewModel.swift`, so no overlap with #1261 (`ShelfView`/`FileShareView`, different regions), #1303, or #1465.

### DIFF
--- a/boringNotch/components/Shelf/ViewModels/ShelfStateViewModel.swift
+++ b/boringNotch/components/Shelf/ViewModels/ShelfStateViewModel.swift
@@ -57,6 +57,24 @@ final class ShelfStateViewModel: ObservableObject {
         items.removeAll { $0.id == item.id }
     }
 
+    /// Removes several items in one pass, so the shelf is persisted once rather than per item.
+    func remove(_ itemsToRemove: [ShelfItem]) {
+        guard !itemsToRemove.isEmpty else { return }
+        let ids = Set(itemsToRemove.map(\.id))
+        for item in itemsToRemove {
+            item.cleanupStoredData()
+        }
+        items.removeAll { ids.contains($0.id) }
+    }
+
+    func removeAll() {
+        guard !items.isEmpty else { return }
+        for item in items {
+            item.cleanupStoredData()
+        }
+        items = []
+    }
+
     func updateBookmark(for item: ShelfItem, bookmark: Data) {
         guard let idx = items.firstIndex(where: { $0.id == item.id }) else { return }
         if case .file = items[idx].kind {
--- a/boringNotch/components/Shelf/Views/ShelfView.swift
+++ b/boringNotch/components/Shelf/Views/ShelfView.swift
@@ -15,6 +15,7 @@ struct ShelfView: View {
     @StateObject var selection = ShelfSelectionModel.shared
     @StateObject private var quickLookService = QuickLookService()
     private let spacing: CGFloat = 8
+    @State private var isHoveringPanel = false
 
     var body: some View {
         HStack(spacing: 12) {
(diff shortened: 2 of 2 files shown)


pcruiher08:
Closing this in favour of agreeing the design on the issue first.

This PR proposed UX that nobody had asked for: #1248 requested a plain Clear All button and a Delete/Backspace shortcut, and what I opened here was a scoped `Remove Selected (n)` / `Clear All` control with no keyboard support. Better to settle the shape before asking anyone to review code.

Since then I have got the keyboard half working as well, with no new permissions required, so the branch now covers both halves of the issue. Implementation, screenshots, the focus mechanism and the one behavioural trade-off are written up here:

https://github.com/TheBoredTeam/boring.notch/issues/1248#issuecomment-5318596960

The branch is ready and CI is green. Happy to reopen or raise a fresh PR once there is a direction. Sorry for the noise.

PR #1488, opened 2026-08-24
Add Dock icon visibility setting

kasyan1337:
## Summary

- Add a default-on “Hide icon in Dock” setting under General → System features.
- Apply the preference at launch and whenever activation policy changes, including while Settings is open.
- Preserve the existing menu-bar-only default while allowing users to show the Dock icon when desired.

## Verification

- Built successfully with Xcode 26.6 using the macOS Debug target.
- Confirmed the built app retains LSUIElement = true.
- Confirmed no existing Dock visibility setting was present in the upstream dev branch.

### DIFF
--- a/boringNotch/components/Settings/SettingsWindowController.swift
+++ b/boringNotch/components/Settings/SettingsWindowController.swift
@@ -67,8 +67,7 @@ class SettingsWindowController: NSWindowController {
     }
     
     func showWindow() {
-        // Set app to regular mode first
-        NSApp.setActivationPolicy(.regular)
+        NSApp.applyDockIconPreference()
         
         // If window is already visible, bring it to front properly
         if window?.isVisible == true {
@@ -100,8 +99,7 @@ class SettingsWindowController: NSWindowController {
     private func relinquishFocus() {
         window?.orderOut(nil)
         
-        // Set app back to accessory mode immediately
-        NSApp.setActivationPolicy(.accessory)
+        NSApp.applyDockIconPreference()
     }
 }
 
@@ -115,8 +113,7 @@ extension SettingsWindowController: NSWindowDelegate {
     }
     
     func windowDidBecomeKey(_ notification: Notification) {
-        // Ensure app is in regular mode when window becomes key
-        NSApp.setActivationPolicy(.regular)
+        NSApp.applyDockIconPreference()
     }
     
     func windowDidResignKey(_ notification: Notification) {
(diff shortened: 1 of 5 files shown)


boringNotch/components/Settings/SettingsWindowController.swift
```diff
@@ -67,8 +67,7 @@ class SettingsWindowController: NSWindowController {
     }
     
     func showWindow() {
-        // Set app to regular mode first
-        NSApp.setActivationPolicy(.regular)
+        NSApp.applyDockIconPreference()
```
CoderXXL, line 70: The `.regular` probably wasn't about the Dock icon. Per the comment in setupWindow() "_Make it behave like a regular app window with proper Spaces support_" and plus `collectionBehavior = [.managed, .participatesInCycle, .fullScreenAuxiliary]`, it was a deliberate choice to give Settings a real app menu and a ⌘-Tab entry. 

so I would leave the shopWindow on .regular and apply the preference only when the window closes. 


```suggestion
        // Set app to regular mode first
        NSApp.setActivationPolicy(.regular)
```

boringNotch/components/Settings/SettingsWindowController.swift
```diff
@@ -115,8 +113,7 @@ extension SettingsWindowController: NSWindowDelegate {
     }
     
     func windowDidBecomeKey(_ notification: Notification) {
-        // Ensure app is in regular mode when window becomes key
-        NSApp.setActivationPolicy(.regular)
+        NSApp.applyDockIconPreference()
```
CoderXXL, line 116: Same as in `showWindow()`. I'd leave this one as .regular.

boringNotch/components/Settings/SettingsWindowController.swift
```diff
@@ -100,8 +99,7 @@ class SettingsWindowController: NSWindowController {
     private func relinquishFocus() {
         window?.orderOut(nil)
         
-        // Set app back to accessory mode immediately
-        NSApp.setActivationPolicy(.accessory)
+        NSApp.applyDockIconPreference()
```
CoderXXL, line 102: 👍🏻

boringNotch/components/Settings/Views/GeneralSettingsView.swift
```diff
@@ -43,6 +44,12 @@ struct GeneralSettings: View {
                     Text("Show menu bar icon")
                 }
                 .tint(.effectiveAccent)
+                Defaults.Toggle(key: .hideDockIcon) {
```
CoderXXL, line 47: worth adding the entry in Localizable

CoderXXL:
Nice feature if you ask me

Issue #1046, opened 2026-02-19
[Bug] issue with signing certificate


hachirotahoshino:
### What happened? What are the steps to reproduce the issue?

Hi, i wanted to add the app to my mac but it ends up with an gatekeeper error, and when i open the app in apparency to check the certificate and notarize status, i end up with a big no:

<img width="1212" height="917" alt="Image" src="https://github.com/user-attachments/assets/74d33a71-56c0-40d8-8abd-4fceb5f9bb2c" />
<img width="416" height="394" alt="Image" src="https://github.com/user-attachments/assets/e0649e29-8cf4-4d49-9a6a-2dd574346d90" />

### What did you expect to happen?

the app to be allowed to run by gatekeepe

### Boring Notch Version

v2.7.3

### macOS Version

26.3

### Music Source? (If relevant)

None

### Relevant log output

```shell
par défaut	07:15:48.995687+0100	lsd	theboringteam.boringnotch/Unknown Persona:5:5:1:1:Building bundle record for app
par défaut	07:15:48.995776+0100	lsd	theboringteam.boringnotch/Unknown Persona:5:5:2:1:Built bundle record for app
par défaut	07:15:48.997442+0100	lsd	theboringteam.boringnotch/Unknown Persona:5:4:2:1:_LSServerRegisterItemInfo result = 0
par défaut	07:15:49.006700+0100	runningboardd	Launch request for app<application.theboringteam.boringnotch.16.21(502)>[0] is using uid 502 (divined from auid 502 euid 502)
par défaut	07:15:49.006743+0100	runningboardd	Acquiring assertion targeting app<application.theboringteam.boringnotch.16.21(502)> from originator [osservice<com.apple.coreservices.uiagent(502)>:17340] with description <RBSAssertionDescriptor| "LS launch theboringteam.boringnotch" ID:467-17340-89723 target:app<application.theboringteam.boringnotch.16.21(502)> attributes:[
	<RBSDomainAttribute| domain:"com.apple.launchservicesd" name:"LaunchRoleUserInteractive" sourceEnvironment:"(null)">
	]>
par défaut	07:15:49.006788+0100	runningboardd	Assertion 467-17340-89723 (target:app<application.theboringteam.boringnotch.16.21(502)>) will be created as active
[...]

PR #1492, opened 2026-08-26
Add missing NSCalendars/NSRemindersFullAccessUsageDescription keys

Andros94:
## Problem

Calendar and Reminders permissions can never be granted on macOS 14+. The app is missing the `FullAccess` usage description keys that EventKit requires.

The code requests **full** access:

```
$ strings -a boringNotch.app/Contents/MacOS/boringNotch | grep requestFullAccess
requestFullAccessToEvents
requestFullAccessToReminders
```

But the shipped `Info.plist` only declares the legacy keys:

```
$ plutil -p boringNotch.app/Contents/Info.plist | grep -iE "Calendars|Reminders"
"NSCalendarsUsageDescription" => "This app uses the calendar to display your calendar events"
"NSRemindersUsageDescription" => "This app uses Reminders to display your scheduled reminder in the calendar"
```

The target is built against the macOS 15.5 SDK (`DTSDKName = macosx15.5`, `LSMinimumSystemVersion = 14.0`). For apps linked on or after macOS 14, `NSCalendarsUsageDescription` is ignored — `requestFullAccessToEvents` requires `NSCalendarsFullAccessUsageDescription`, and `requestFullAccessToReminders` requires `NSRemindersFullAccessUsageDescription`.

The result on v2.7.3: the access request fails silently, **no permission prompt is ever shown**, and the app never appears in System Settings → Privacy & Security → Calendars. The calendar view stays empty even with fully synced accounts in Calendar.app.

Only `INFOPLIST_KEY_NSCalendarsUsageDescription` and `INFOPLIST_KEY_NSRemindersUsageDescription` are set in `project.pbxproj`, so the required keys never reach the built bundle.

## Fix

Add `INFOPLIST_KEY_NSCalendarsFullAccessUsageDescription` and `INFOPLIST_KEY_NSRemindersFullAccessUsageDescription` to both the Debug and Release configurations, reusing the existing descriptions.

The legacy keys are kept, so the change is purely additive. Four added lines, no code changes and no translation changes.

## Testing

`xcodebuild -scheme boringNotch -configuration Debug` → **BUILD SUCCEEDED**, and the resulting bundle now carries all four keys:

```
$ plutil -p Build/Products/Debug/boringNotch.app/Contents/Info.plist | grep -iE "Calendars|Reminders"
"NSCalendarsFullAccessUsageDescription" => "This app uses the calendar to display your calendar events"
"NSCalendarsUsageDescription" => "This app uses the calendar to display your calendar events"
"NSRemindersFullAccessUsageDescription" => "This app uses Reminders to display your scheduled reminder in the calendar"
"NSRemindersUsageDescription" => "This app uses Reminders to display your scheduled reminder in the calendar"
```

No screenshots: this changes build settings only, there is no UI change.

Reproduced on macOS 26.0 (Darwin 25.6.0), MacBook Air M2, boringNotch 2.7.3.

### DIFF
--- a/boringNotch.xcodeproj/project.pbxproj
+++ b/boringNotch.xcodeproj/project.pbxproj
@@ -1350,9 +1350,11 @@
 				INFOPLIST_KEY_LSBackgroundOnly = NO;
 				INFOPLIST_KEY_LSUIElement = YES;
 				INFOPLIST_KEY_NSAppleEventsUsageDescription = "This app uses AppleEvents to control music";
+				INFOPLIST_KEY_NSCalendarsFullAccessUsageDescription = "This app uses the calendar to display your calendar events";
 				INFOPLIST_KEY_NSCalendarsUsageDescription = "This app uses the calendar to display your calendar events";
 				INFOPLIST_KEY_NSCameraUsageDescription = "This app uses the camera to display a live camera view";
 				INFOPLIST_KEY_NSHumanReadableCopyright = "";
+				INFOPLIST_KEY_NSRemindersFullAccessUsageDescription = "This app uses Reminders to display your scheduled reminder in the calendar";
 				INFOPLIST_KEY_NSRemindersUsageDescription = "This app uses Reminders to display your scheduled reminder in the calendar";
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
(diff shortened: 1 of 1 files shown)


Alexander5015:
I don't think there are any issues with the calendar and reminder permissions right now. While it is true that the older keys are in use right now, I haven't seen any reports of this causing any issues, nor have I had any trouble using either myself. Could you provide more info on the issue with the calendar not working?

Andros94:
You're right, and I was wrong — sorry for the noise.

I tested this properly instead of relying on the documented requirement. I built two minimal app bundles against the macOS 26.5 SDK, identical except for their `Info.plist` keys, each calling `requestFullAccessToEvents`. With **only** `NSCalendarsUsageDescription`:

```
cles plist  : ["NSCalendarsUsageDescription"]
statut avant: notDetermined
--> appel requestFullAccessToEvents
granted     : true
error       : aucune
statut apres: fullAccess
```

The permission prompt appears normally and the call succeeds. macOS accepts the legacy key as a fallback for the full-access request, so the current keys are fine and this PR fixes nothing.

My original symptom had a different cause. The app was being SIGKILLed by Gatekeeper roughly 4 seconds after launch because of the `com.apple.quarantine` attribute on the non-notarized build — no crash report, just a silent exit 137. It never lived long enough to request calendar access, which is why it never appeared under Privacy & Security → Calendars. Running `xattr -dr com.apple.quarantine /Applications/boringNotch.app` resolved it, and the prompt then showed up as expected. That's the known signing situation from #1046, not a permissions bug.

Closing this. Thanks for pushing back rather than just merging it.

Issue #1054, opened 2026-02-25
Move project to Swift 6


Alexander5015:
### Summary
Track the migration of the codebase to Swift 6, including language mode adoption, concurrency updates, dependency compatibility, and CI adjustments.

Swift 6 introduces stricter concurrency checking, enhanced Sendable enforcement, and changes that may surface previously ignored warnings as errors. This issue covers the required work to ensure full compatibility and a clean build under Swift 6 language mode.
### Motivation
- Align with the latest Swift language evolution
- Maintain compatibility with the current Apple toolchain
- Improve concurrency safety with stricter compile-time checks
- Reduce technical debt by addressing warnings that will become errors

Dbhardwaj99:
I'd like to pick this up if no one's on it.

Current state on `dev` as far as I can tell: `SWIFT_VERSION` is still 5.0 across all six
build configs, and `SWIFT_STRICT_CONCURRENCY` is `targeted` on the app and XPC helper
configs but unset at the project level. The existing debt looks small — one
`@unchecked Sendable` (`helpers/AssociatedObject.swift`), one `@preconcurrency` import
(`Providers/CalendarServiceProviding.swift`), no `nonisolated(unsafe)` — though
`DispatchQueue` is still used across ~23 files, with `WebcamManager`,
`AudioCaptureManager`, `MusicManager` and `NSImage+Extensions` the heaviest.

Since a full migration touches most of the codebase, I don't think one PR is the right
shape — it would conflict badly with the current queue, #1401 especially. I'd suggest:

1. Set `SWIFT_STRICT_CONCURRENCY = complete` while staying in Swift 5 language mode. No
   behaviour change, but it turns the remaining work into a countable list of warnings.
2. Clear those warnings in separate PRs — XPC helper target first (the private-framework
   handles and XPC callbacks there look like the tricky part), then managers, then views.
3. Flip `SWIFT_VERSION` to 6.0 on both targets once it's clean.

Does that split work, or would you rather have it in fewer, larger steps? I can start
with step 1 so there's a concrete number to plan against.

Alexander5015:
[...]

PR #1495, opened 2026-08-28
Isolate XPCHelperClient to the main actor

Dbhardwaj99:
Second step of the Swift 6 migration tracked in #1054, following #1430.

## The problem

`XPCHelperClient` kept its connection state (`remoteService`, `connection`, `lunarListener`,
`monitoringTask`) as mutable properties on a non-isolated class, while every method reached that
state through `await MainActor.run { ensureRemoteService() }` and then used the returned service
outside the actor.

That sent two things across an isolation boundary on every call, which is where 56 of the
project's strict-concurrency diagnostics came from:

| Count | Diagnostic |
|------:|------------|
| 19 | `sending 'self' risks causing data races` |
| 19 | `capture of 'self' with non-Sendable type 'XPCHelperClient' in a '@Sendable' closure` |
| 15 | `type 'RemoteXPCService<any BoringNotchXPCHelperProtocol>' does not conform to 'Sendable'` |
| 2 | `passing closure as a 'sending' parameter` |
| 1 | `static property 'shared' is not concurrency-safe` |

## The change

Isolating the class to the main actor removes the boundary instead of annotating each crossing.
The `MainActor.run` hops go away, so neither `self` nor the service ever leaves the actor, and the
type becomes implicitly `Sendable`.

- `ensureRemoteService()` is called directly rather than through `MainActor.run`
- `startMonitoringAccessibilityAuthorization` uses `Task` instead of `Task.detached` so it stays
  on the actor
- `shared` and `init` remain `nonisolated`, which is safe now that the type is `Sendable`. This
  keeps the existing call contract, so non-isolated callers such as `MediaKeyInterceptor` compile
  unchanged
- `requestAccessibilityAuthorization()` remains `nonisolated` because it is fire and forget, and
  its work hops onto the actor internally
- the `deinit` is removed and `init` made private. A main-actor class has a non-isolated `deinit`
  that cannot touch isolated stored properties, and for a process-lifetime singleton it was
  unreachable regardless. Teardown is already explicit in `applicationWillTerminate`, which calls
  `stopMonitoringAccessibilityAuthorization()` directly

Net effect on the file is 43 insertions and 80 deletions.

## Results

Project-wide diagnostics that the compiler marks as `this is an error in the Swift 6 language
mode`, measured on clean Debug builds before and after:

| | Before | After |
|---|---:|---:|
| **Project total** | 227 | **164** |
| `XPCHelperClient.swift` | 56 | **0** |
| `components/OSD/Managers/XPC/BrightnessManager.swift` | 18 | 10 |
| `components/OSD/Managers/LunarManager.swift` | 13 | 14 |

`BrightnessManager` improves without being touched, because it calls into this client.

`LunarManager` gains one: `sending value of non-Sendable type 'LunarEventListener'`. Now that
`startLunarEventStream(listener:)` is main-actor isolated, passing a non-Sendable listener into it
from a detached task is a crossing. That is a pre-existing problem in `LunarManager` which this
change surfaces rather than causes, and that file is already on the list for its own pass.

One practical note for planning: per-file counts shift as these land, so the remaining split is
worth re-measuring before each PR rather than working from the original table in #1054.

## Testing

Tested on a locally signed build, since the ad-hoc signed helper in a stock Debug build can't be
granted Accessibility. That signing change isn't part of this PR.

Accessibility prompt works, brightness, volume and keyboard backlight all show in the notch and
take effect, toggling Replace System OSD off and on restarts the monitor fine, and quitting from
the menu exits cleanly. Didn't test the Lunar paths since I don't have Lunar installed.

Relates to #1054. Further steps remain.

### DIFF
--- a/boringNotch/XPCHelperClient/XPCHelperClient.swift
+++ b/boringNotch/XPCHelperClient/XPCHelperClient.swift
@@ -2,9 +2,14 @@ import Foundation
 import Cocoa
 import AsyncXPCConnection
 
+@MainActor
 final class XPCHelperClient: NSObject {
     nonisolated static let shared = XPCHelperClient()
     
+    nonisolated private override init() {
+        super.init()
+    }
+
     private let serviceName = "theboringteam.boringnotch.BoringNotchXPCHelper"
     
     private var remoteService: RemoteXPCService<BoringNotchXPCHelperProtocol>?
@@ -14,14 +19,8 @@ final class XPCHelperClient: NSObject {
     private var lunarListener: BoringNotchXPCHelperLunarListener?
     private var hasLunarListener: Bool = false
     
-    deinit {
-        connection?.invalidate()
-        stopMonitoringAccessibilityAuthorization()
-    }
-    
     // MARK: - Connection Management (Main Actor Isolated)
     
-    @MainActor
     private func ensureRemoteService(needsListener: Bool = false) -> RemoteXPCService<BoringNotchXPCHelperProtocol> {
         if let existing = remoteService, (!needsListener || hasLunarListener) {
             return existing
@@ -72,7 +71,6 @@ final class XPCHelperClient: NSObject {
         return service
     }
     
-    @MainActor
     private func getRemoteService() -> RemoteXPCService<BoringNotchXPCHelperProtocol>? {
         remoteService
     }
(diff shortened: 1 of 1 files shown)


theboringhumane:
fixed in #1496

PR #1500, opened 2026-08-29
Add Apple Notes integration

AustinPerfect:
## Summary

- Add a Notes tab with an expandable search field, color filters, pinned note cards, clipboard creation, and list-to-editor transitions.
- Import unlocked notes from every Apple Notes account through Apple Events.
- Save notes created in the app to a dedicated Boring Notch folder and allow those managed notes to be edited or deleted.
- Keep pre-existing Apple Notes content read-only in the editor while still allowing local pinning and color labels for every imported note.
- Add a setting that lets users hide pre-existing Apple Notes and show only notes created by Boring Notch.
- Add English, Simplified Chinese, and Traditional Chinese localizations for the Notes experience.
- Add parser coverage for malformed records, managed-note state, presentation metadata, dates, untitled notes, and imported-note filtering.

## Sync and safety

The integration tracks IDs for notes created by Boring Notch and restricts content updates and deletion to those IDs. Existing notes can be opened in Apple Notes for native editing. Local pin and color metadata are stored by Apple Note ID and do not rewrite the note body.

This boundary avoids flattening rich Apple Notes content into plain text while the structured body format is not yet fully supported. Password-protected notes are skipped.

## Notes structure findings

A read-only inspection of two test notes through the public Notes scripting interface produced the following results without modifying or printing their content:

- The scripting dictionary exposes the note body as writable HTML, plaintext as read-only text, and attachments as separate note elements.
- Each inspected body exposed two table and tbody elements, four table rows, and eight table cells.
- Both bodies exposed unordered-list and list-item elements.
- The attachment collections contained two and three attachments respectively.
- Embedded image data was present through data URL sources together with object elements.
- No checkbox, checked attribute, checkbox symbol, or Apple-specific checklist marker appeared in either the HTML body or plaintext export.

These findings indicate that table and basic list rendering can be implemented from the exported HTML. Reliable checklist state still requires a controlled checked and unchecked checklist sample. Structured table, checklist, and attachment editing is intentionally deferred to a follow-up change so this pull request remains focused on the Notes integration and safe plain-text workflow.

## Interaction fixes

- Resize the panel when switching to the Notes list or editor and preserve the rounded bottom corners.
- Ignore false hover exits at the top screen edge to prevent repeated open and close cycles.
- Suppress the global close gesture while scrolling inside Notes.
- Allow the panel to become key, accept the first mouse click, and focus the new-note editor so text input works.

## Testing

- macOS arm64 Debug build with code signing disabled
- Full boringNotch test suite
- String catalog JSON validation
- Git whitespace validation

The branch is based directly on dev and does not include the lock screen widget commits or files.


AustinPerfect:
hi

AustinPerfect:
Anyone？

PR #1496, opened 2026-08-28
Notification live activity, compact mode & audit remediation (phases 0-4)

theboringhumane:
## Summary

67 commits rebased onto current `dev`. Three bodies of work:

### Notification live activity
- XPC helper captures system banners via Accessibility; chin pill on the closed notch shows the sender app icon + live dot, with a dedicated OTP code pill
- Expanded card on manual open: working reply field (held banners, keystroke fixes, send/hand-off sounds), iMessage replies via Messages scripting, WhatsApp reply drafting in-conversation, Apple Intelligence smart-reply suggestions, per-stage delivery timeouts
- Queueing for notifications arriving mid-reply, with a browsable stack badge
- Reliability fixes: WhatsApp drops, XPC callback/connection liveness, watcher startup scanning, stale-notification hijacking, expiry-while-opening, held-banner leaks, AXI watcher memory pressure
- **Sender icon resolution hardened** — on-disk fallbacks in the helper + app-side re-resolution, so surfaces show the real app icon instead of a grey bell
- **Sneak peek removed entirely** — notifications surface only as the chin pill and the expanded card on manual open

### Compact mode
- Player-only opened notch matching Atoll's layout and dimensions (336pt), full transport row, media output device switching, shuffle/output in both layouts, cached track state instead of "Nothing Playing"

### Audit remediation (phases 0-4)
- Crash fixes, dead-code purge, OSS compliance; background polling killed, OSD hot paths fixed; coordinator cycle broken, architecture decoupled; naming/dedup/conventions/structured logging; community files, tests, lint, security docs; app-target DEVELOPMENT_TEAM restored

## Test plan
- [x] `xcodebuild -scheme boringNotch build` — clean
- [x] 13 targeted tests pass: `NotchUIEventTests`, `PlaybackStateTests`, `ExpandedViewPixelTests` (expanded card paints content), `BundleIDResolverTests` (8)
- [ ] Manual: receive a banner → chin pill shows app icon + live dot → open notch → expanded card → send a reply

PR #1503, opened 2026-08-29
Notification live activity on the notch [1/4]

theboringhumane:
**Stack 1 of 4.** Base for #2. Split out of #1496 for reviewability.

## Summary
The notification system foundation: an XPC helper captures system notification banners via Accessibility and the notch mirrors them.

- **Closed notch**: chin pill with sender app icon + live dot; dedicated OTP code pill with tap-to-copy
- **Open notch**: expanded card — reply field (keystroke/focus fixes, dismiss pause while typing, keep-open while composing), Apple Intelligence smart-reply suggestions, iMessage replies via Messages scripting, send/hand-off sounds
- **Queueing**: notifications arriving mid-reply are queued; dismiss/expiry lifecycle managed
- **Reliability**: WhatsApp drop fix, XPC callback/connection liveness, watcher startup scanning, stale-notification hijacking fix, banner hold+off-screen parking so replies keep working

## Test plan
- [x] `xcodebuild -scheme boringNotch build` at this tip — clean
- [ ] Manual: banner → chin pill → open → reply

### DIFF
--- a/BoringNotchXPCHelper/MessagesSender.swift
+++ b/BoringNotchXPCHelper/MessagesSender.swift
@@ -1,0 +1,85 @@
+//
+//  MessagesSender.swift
+//  BoringNotchXPCHelper
+//
+//  Sends iMessage replies through the Messages scripting dictionary.
+//
+//  Why this exists: replying to a notification normally means typing into
+//  the system banner's own AX reply field, which stops working the moment
+//  that banner fades (~5s) — the element is destroyed, measured, not
+//  assumed. Messages is the one supported app with a real scripting
+//  dictionary (`send <text> to <chat|participant>`), so an iMessage reply
+//  can be delivered properly at any time, with no dependency on the
+//  notification still existing.
+//
+//  There is no equivalent for WhatsApp/Telegram/Discord — none ship a
+//  scripting dictionary, and driving their UI with simulated clicks is far
+//  too brittle to put behind a send button. Those keep the clipboard
+//  hand-off.
+//
+//  Lives in the helper because the app is sandboxed; sending Apple events
+//  to another app needs the unsandboxed context the helper already has.
+//
+
+import Foundation
+import AppKit
+
+enum MessagesSender {
+    /// Matches the notification's sender against a Messages chat (then a
+    /// participant) and sends. Returns false if the chat can't be resolved
+    /// or automation permission was denied, so the caller can fall back to
+    /// the clipboard hand-off.
+    static func send(_ text: String, toChatNamed name: String) -> Bool {
+        // Participants only — deliberately not chats. Verified against a
+        // real Messages library: `name of chat` returns `missing value` for
+        // every chat, so matching on it can never succeed. Participants do
+        // carry the display name the notification shows ("Harsh Vardhan
+        // Goswami"), which is the only thing a notification gives us.
+        //
+        // The first match wins. Duplicate participant entries for the same
+        // person are normal (one per handle/service — e:me@…, +9198…), and
+        // they all reach the same human, so picking the first is fine.
+        let script = """
+        tell application "Messages"
+            repeat with p in participants
+                try
+                    if (name of p as string) is equal to "\(escape(name))" then
+                        send "\(escape(text))" to p
+                        return "ok"
+                    end if
+                end try
+            end repeat
+        end tell
+        return "notfound"
+        """
+
+        guard let scriptObject = NSAppleScript(source: script) else { return false }
+
+        var error: NSDictionary?
+        let output = scriptObject.executeAndReturnError(&error)
+
+        if let error {
+            // -1743 is "not authorized to send Apple events" — the user
+            // declined the Automation prompt, which is a legitimate choice,
+            // not a bug. Everything else is worth seeing in the log.
+            NSLog("[boringNotch] Messages send failed: \(error)")
+            return false
+        }
+
+        let ok = output.stringValue == "ok"
+        if !ok {
+            NSLog("[boringNotch] Messages: no chat or participant named \(name.debugDescription)")
+        }
+        return ok
+    }
+
+    /// Message text is arbitrary user input going into an AppleScript
+    /// string literal, so backslashes and quotes have to be neutralised —
+    /// otherwise a quote in a reply breaks the script (or worse, changes
+    /// what it does). Backslash first, or it would re-escape the escapes.
+    private static func escape(_ value: String) -> String {
+        value
+            .replacingOccurrences(of: "\\", with: "\\\\")
+            .replacingOccurrences(of: "\"", with: "\\\"")
+    }
+}
--- a/BoringNotchXPCHelper/NotificationWatcher.swift
+++ b/BoringNotchXPCHelper/NotificationWatcher.swift
@@ -150,19 +150,25 @@ final class NotificationWatcher {
         }
     }
 
-    /// Holds a banner open so its reply field stays usable, optionally
-    /// moving it off-screen first so the user never sees it.
+    /// Holds a banner open so its reply field stays usable, moving it
+    /// off-screen first.
     ///
-    /// The off-screen move is safe to leave behind: when no banners are
-    /// showing, notificationcenterui has zero windows — the banner window is
-    /// created per session and destroyed after — so a moved window can't
-    /// permanently hide notifications. A fresh one spawns at its normal
-    /// position.
-    func hold(token: String, offScreen: Bool) {
+    /// The move is not optional. Holding a banner works by re-performing
+    /// its details toggle, which leaves it expanded — showing its own reply
+    /// field, on top of everything, taking focus. Two live text fields
+    /// competing for the same keystrokes is worse than no keep-alive at
+    /// all, so the banner is parked off-screen for the whole hold and the
+    /// notch is the only visible surface.
+    ///
+    /// Safe to leave behind: with no banners showing, notificationcenterui
+    /// has zero windows — the window is created per session and destroyed
+    /// after — so a moved window can't permanently hide notifications. A
+    /// fresh one always spawns at its normal position.
+    func hold(token: String) {
         guard let banner = live[token] else { return }
         held.insert(token)
 
-        guard offScreen, !heldOffScreen.contains(token) else { return }
+        guard !heldOffScreen.contains(token) else { return }
         heldOffScreen.insert(token)
         guard let windowValue = banner[kAXWindowAttribute] else { return }
         let window = windowValue as! AXUIElement
--- a/BoringNotchXPCHelper/NotificationWatcher.swift
+++ b/BoringNotchXPCHelper/NotificationWatcher.swift
@@ -39,10 +39,13 @@ final class NotificationWatcher {
     var onBanner: ((CapturedNotification) -> Void)?
     var onBannerGone: ((String) -> Void)?
 
-    private var observer: AXObserver?
     private var appElement: AXUIElement?
-    private var pollTimer: Timer?
+    private var pollTimer: DispatchSourceTimer?
     private var live: [String: AXUIElement] = [:]
+
+    /// Banners live ~5s, so this catches every one with room to spare while
+    /// staying cheap — each tick is a shallow AX tree walk.
+    private let pollInterval: TimeInterval = 0.35
 
     var isRunning: Bool { appElement != nil }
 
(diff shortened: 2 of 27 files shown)


BoringNotchXPCHelper/NotificationWatcher.swift
```diff
@@ -0,0 +1,446 @@
...
+    /// has zero windows — the window is created per session and destroyed
+    /// after — so a moved window can't permanently hide notifications. A
+    /// fresh one always spawns at its normal position.
+    func hold(token: String) {
+        guard let banner = live[token] else { return }
+        held.insert(token)
+
+        guard !heldOffScreen.contains(token) else { return }
+        heldOffScreen.insert(token)
+        guard let windowValue = banner[kAXWindowAttribute] else { return }
```
Alexander5015, line 173: Correct me if I’m wrong, but this per-banner hold appears to move Notification Center’s entire shared window off-screen, not the individual banner. `scan()` can find multiple banners under one window, but `hold(token:)` follows the banner’s `kAXWindowAttribute` and moves that window to `(-5000, -5000)`. That means unrelated or filtered notifications in the same window could disappear too, and the per-token state does not track the window’s original position or restore it when watching stops. I do not think window manipulation is safe here.

boringNotch/managers/SystemNotificationManager.swift
```diff
@@ -0,0 +1,364 @@
...
+        // against the tail of an allowed bundle ID — "WhatsApp" against
+        // net.whatsapp.WhatsApp.
+        guard let appName = notification.appName?.lowercased(), !appName.isEmpty else { return false }
+        return Defaults[.notificationAllowedApps].contains { bundleID in
+            bundleID.split(separator: ".").last.map { appName == $0.lowercased() } ?? false
+        }
+    }
+
+    private func show(_ notification: SystemNotification) {
+        withAnimation(.smooth) { activeNotification = notification }
```
Alexander5015, line 202: I don't think this queues notifications the way the PR description implies. Every arrival immediately overwrites `activeNotification`, while the draft and composing/sending state live in view-local `@State` and `.id(notification.id)` rebuilds that view. If B arrives while A is being written, I think A’s draft is wiped out, and because `dismissActive` only releases the current notification, A’s AX target can be orphaned.

BoringNotchXPCHelper/MessagesSender.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// or automation permission was denied, so the caller can fall back to
+    /// the clipboard hand-off.
+    static func send(_ text: String, toChatNamed name: String) -> Bool {
+        // Participants only — deliberately not chats. Verified against a
+        // real Messages library: `name of chat` returns `missing value` for
+        // every chat, so matching on it can never succeed. Participants do
+        // carry the display name the notification shows ("Harsh Vardhan
+        // Goswami"), which is the only thing a notification gives us.
+        //
+        // The first match wins. Duplicate participant entries for the same
```
Alexander5015, line 39: This fallback won't reply to the originating Messages conversation. It ignores chats and sends to the first global participant whose display name matches the notification title. It also will not know the account, service type, and handle, so it does not preserve iMessage/SMS/RCS routing. This can choose the wrong person, thread, handle, or transport and still return “ok”.

BoringNotchXPCHelper/NotificationWatcher.swift
```diff
@@ -0,0 +1,446 @@
...
+    /// banner's actions are ["AXPress", "Show Details", "Reply", "Close"] —
+    /// a real "Reply" action does exist (an earlier check only ever caught
+    /// the banner post-expansion, where it's already gone, which led to a
+    /// wrong assumption here). "Reply" is preferred since it's the more
+    /// direct, purpose-built action; "Show Details" is the fallback for
+    /// banners/apps that don't expose it. Either way the field is submitted
+    /// via the "Send" action on the banner — not kAXConfirmAction on the
+    /// field, which is untested and unreliable across text-area
+    /// implementations.
+    func reply(token: String, text: String) -> Bool {
```
Alexander5015, line 348: This is too brittle to treat as a confirmed send operation. It drives a foreign, asynchronous AX hierarchy, selects Reply/Details/Send through English substrings observed in one environment, chooses the first text-field descendant, and then sleeps for exactly 400 ms on the helper’s main queue. That blocks polling and hold refresh while still failing if the field appears later. `AXUIElementPerformAction == .success` only establishes that the AX request succeeded, not that the message was delivered, but the caller reports it as sent and plays a confirmation sound.

Alexander5015 (commented):
Only had time to review a few files, here are my current comments.

BoringNotchXPCHelper/NotificationWatcher.swift
theboringhumane: You're right on both counts, and we've reworked it (bc4906d, landed on stack tip #1506 since the branches build on each other): (1) the park is now gated on the notch being open — the same gate as the keep-alive expand — so arrivals with a closed notch are held in state only and never touch the window at all (they expire naturally, as before the hold feature); parking happens via holdActive() when the user actually opens the notch. (2) The window's original position is now recorded on first park, keyed by CFHash(window), and restored when the last held token referencing that window releases, expires, or watching stops — and if the origin can't be read, we skip parking entirely rather than make an unrestorable move.

boringNotch/managers/SystemNotificationManager.swift
theboringhumane: Fair at this snapshot — this is exactly what the later stack commits add. In #1504, 877dbf4 ("Queue notifications that arrive mid-reply instead of replacing") gates replacement behind isComposingReply and enqueues B instead of overwriting A, with a browsable stack badge; and drafts are held manager-side keyed by notification id (replyDrafts in SystemNotificationManager), not view-local @State, so the .id rebuild no longer wipes them. On the orphaned AX target: show()'s releasingPrevious path explicitly releases the superseded notification's hold before replacing it, and dd0d381 ("Fix held banner leak that could block the real Notification Center", also #1504) closes the remaining leak. Happy to pull the queueing commit forward into this PR if you'd rather review it here.

BoringNotchXPCHelper/MessagesSender.swift
theboringhumane: Agreed, and fixed (d1ae121, on stack tip #1506): the script now resolves chats first — it matches a chat's participants by display name (chat name itself is missing value for 1:1 threads, which is what sent us to participants originally), prefers a single-participant chat, and sends INTO the chat object, so account, service (iMessage/SMS/RCS), handle, and thread are all preserved. The bare-participant path is now a guarded fallback: it counts matches library-wide and only sends when exactly one participant matches; zero or 2+ (the duplicate-person-across-handles case) returns false and the caller falls back to the clipboard hand-off. Nothing sends on ambiguity anymore.

BoringNotchXPCHelper/NotificationWatcher.swift
theboringhumane: Agreed on all three points, fixed (bc4906d, on stack tip #1506): the reply now runs on a dedicated serial queue — the helper's main queue never blocks, so polling and hold refresh keep running; the fixed 400ms sleep is replaced by a bounded wait that polls for the reply field in 50ms slices up to 1.2s (false on timeout); and after the send action, a bounded verification poll (~1s) only reports success once the field clears or is destroyed with the sent banner's collapse — send-action failure or a non-clearing field reports false. To your delivery point: AX can verify the banner accepted the reply, not network delivery, and the code now says exactly that — the documented limitation stands, but we no longer report "sent" on AX success alone. Field selection also now prefers the AX-focused field over first-in-document-order.

theboringhumane:
Heads up for review continuity: the stack was just rebased onto current `dev` (which merged #1477 — the conflict this PR showed on `Localizable.xcstrings` is resolved). Same 27 commits, resolutions preserved from your own rebase of this branch (`c30a3db`) — the only delta from it is the #1477 merge-in plus your 6-line pbxproj adjustment. New SHAs for the review-fix commits referenced in my replies above: park-gating + reply hardening `bc4906d` → `affdd12`, MessagesSender chat-first routing `d1ae121` → `b948de5` (both on #1506). One resolution worth your eye in #1505: `MediaEnvironment.swift` was deleted — #1477 removed the deprecation-probe API it wrapped, and the old "re-show onboarding on NowPlaying-deprecated" branch is gone; the substrate's passive `nowPlayingNotice` fallback replaces it. Flagging in case you want the onboarding re-prompt re-implemented on top of `NowPlayingAvailability`.

BoringNotchXPCHelper/NotificationWatcher.swift
Alexander5015: Restoring the position doesn't address the main concern: we're still moving the whole Notification Center window, including unrelated notifications the user may have excluded. I don't think we should manipulate that shared window to make this feature work.

The opening path also doesn't match the explanation. [`holdActive()` only cancels the dismiss timer](https://github.com/TheBoredTeam/boring.notch/blob/80b00f69e16b41e5654709ef7a9f0b442912f325/boringNotch/managers/SystemNotificationManager.swift#L381), and [the `notchOpen` setter](https://github.com/TheBoredTeam/boring.notch/blob/80b00f69e16b41e5654709ef7a9f0b442912f325/BoringNotchXPCHelper/NotificationWatcher.swift#L69) doesn't park existing banners. A notification received while closed can therefore be expanded by the next keep-alive in its normal visible position. I haven't tested the focus behavior live, but the claimed parking call isn't there.

boringNotch/managers/SystemNotificationManager.swift
Alexander5015: The queue and draft storage are there now, but the view still carries state between notifications. [`.id` is inside `NotificationExpandedView.body`](https://github.com/TheBoredTeam/boring.notch/blob/80b00f69e16b41e5654709ef7a9f0b442912f325/boringNotch/components/Notch/NotificationLiveActivity.swift#L148), so changing the notification preserves `replyText` and the sending/error state. The appear/disappear handlers don't run either.

A small SwiftUI test with the same structure reproduced B rendering with A's draft and no lifecycle callbacks. Pressing Send would then reply to B with A's text.

Move `.id` to [the `NotificationExpandedView(...)` call in the parent](https://github.com/TheBoredTeam/boring.notch/blob/80b00f69e16b41e5654709ef7a9f0b442912f325/boringNotch/ContentView.swift#L556), and test cycling while typing or sending, including A's send completing after switching to B. Manager-side draft storage alone doesn't fix this.

BoringNotchXPCHelper/MessagesSender.swift
Alexander5015: I don't think this fixes the routing issue. [Both chat paths send to the first match](https://github.com/TheBoredTeam/boring.notch/blob/80b00f69e16b41e5654709ef7a9f0b442912f325/BoringNotchXPCHelper/MessagesSender.swift#L73) before reaching the ambiguity check. If two chats match the name, how do we know which one produced the notification? A group member's direct chat isn't necessarily the right destination either.

Preserving the selected chat's account and service doesn't establish that we selected the right chat. We need an identifier for the actual conversation. If all we have is a display name, we should open Messages and let the user choose where to reply. I wouldn't ship automatic sends based on this lookup.

BoringNotchXPCHelper/NotificationWatcher.swift
Alexander5015: Moving the AX work off main and replacing the fixed sleep addresses those points, but the timeout still doesn't work.

[`raceTimeout`](https://github.com/TheBoredTeam/boring.notch/blob/80b00f69e16b41e5654709ef7a9f0b442912f325/boringNotch/managers/SystemNotificationManager.swift#L460) waits for its child tasks before returning, and cancelling the XPC task doesn't resume its checked continuation. In a synthetic test with no XPC call or message send, a 50 ms deadline returned `false` after about 530 ms, once the simulated callback arrived.

A missing callback can leave the UI waiting. A late successful send is reported as failed, leaving the draft ready to retry. The timeout needs to return independently and distinguish an unknown outcome from failure to avoid encouraging duplicate replies.

[The Messages fallback also still runs synchronous AppleScript on the helper's main queue](https://github.com/TheBoredTeam/boring.notch/blob/80b00f69e16b41e5654709ef7a9f0b442912f325/BoringNotchXPCHelper/BoringNotchXPCHelper.swift#L142), so it can block polling. The English-label matching still needs tests across the apps and macOS versions we support.

PR #1505, opened 2026-08-29
Audit remediation, phases 0-4 [3/4]

theboringhumane:
**Stack 3 of 4.** Based on #1504; base for #1506. Split out of #1496 for reviewability.

## Summary
- **Phase 0**: crash fixes, dead-code purge (incl. `ProgressIndicator.swift`), OSS compliance
- **Phase 1**: background polling killed, OSD hot paths fixed
- **Phase 2**: coordinator retain cycle broken, architecture decoupled
- **Phase 3**: naming, dedup, conventions, structured logging (`Log.swift`); `Live activities/` → `LiveActivities/`
- **Phase 4**: community files, tests, lint, security docs; app-target `DEVELOPMENT_TEAM` restored for automatic signing

Large but mechanical; the feature stacks (#1503, #1504) are refactored here, not changed behaviorally.

## Review round 1 (Alexander5015) — all 12 addressed
- `SECURITY.md`: supported-versions table removed
- `.gitignore`: reverted wholesale (the cleanup over-reached and let an xcscheme slip in); scheme file untracked, `*.xcscheme` ignored again
- `.swiftlint.yml`: narrative comments trimmed; **SwiftLint CI added** (`.github/workflows/swiftlint.yml`, PRs to dev + stack/**)
- `dependabot.yml`: `swift` ecosystem **restored** — the "Dependabot can't read the manifest" claim was wrong (#1434 proves it works)
- `NotchHomeView` slider: interval 0.5s → **0.2s** (reviewer cap) with corrected rationale comment
- `WebcamPreview` → **`WebcamView`** (View-suffix convention, matches filename)
- `SystemEventIndicatorModifier`: narrating comment removed
- `CODEOWNERS`: kept per repo-owner decision — routing, not territory; reply on the thread, happy to reshape the mapping

## Notes
- Rebased over #1477 (Now Playing reliability): availability substrate kept wholesale, audit intent re-applied where valid; `MediaEnvironment.swift` deleted (dead on the new substrate — its deprecation-probe API is gone).

## Test plan
- [x] `xcodebuild -scheme boringNotch build` at this tip — clean
- [x] Stack tip (#1506) builds + 13/13 targeted tests pass with bare commands


SECURITY.md
```diff
@@ -1,5 +1,15 @@
 # Security Policy
 
+## Supported Versions
+
+Only the latest release (and the `main` branch) receives security fixes.
+Beta builds on the `dev` branch are development snapshots.
+
+| Version | Supported |
```
Alexander5015: This table should be removed

boringNotch/components/Notch/NotchHomeView.swift
```diff
@@ -458,7 +460,7 @@ struct NotchHomeView: View {
             }
 
             if shouldShowCamera {
-                CameraPreviewView(webcamManager: webcamManager)
```
Alexander5015, line 461: Typically its best practice to append View to the names of views

.gitignore
Alexander5015, line 1: What is the reason for these changes?

.swiftlint.yml
```diff
@@ -0,0 +1,85 @@
+# SwiftLint configuration for boring.notch
+# Run with: swiftlint --config .swiftlint.yml
+# CI wiring is tracked as a follow-up (needs a lint-capable runner).
```
Alexander5015: Shouldn't be too hard to add this into CI?

.swiftlint.yml
```diff
@@ -0,0 +1,85 @@
...
+
+# Paths to check
```
Alexander5015: Unnecessary comment

.swiftlint.yml
```diff
@@ -0,0 +1,85 @@
...
+included:
+  - boringNotch
+  - BoringNotchXPCHelper
+  - Shared
+
+# Paths to ignore
```
Alexander5015: Unnecessary comment

.swiftlint.yml
```diff
@@ -0,0 +1,85 @@
...
+  severity: error
+  excluded:
+    # XPC secure-coding bridging (NSSet -> Set<AnyHashable> is a guaranteed bridge)
+    - boringNotch/XPCHelperClient/XPCHelperClient.swift
+    - BoringNotchXPCHelper/main.swift
+
+force_try:
+  severity: error
+
+# Opt-in rules matching the house style established during remediation
```
Alexander5015: This is probably unnecessary comment

.github/CODEOWNERS
Alexander5015, line 1: I don't think we need this, I don't think it aligns well given that these PRs make significant changes to the Area-specific expertise you carved out, which means you should have a great understanding of those areas, and I have made significant changes throughout the codebase, so I also wouldn't consider you more of an expert in every other area either.

.github/dependabot.yml
```diff
@@ -5,18 +5,20 @@
...
     schedule:
       interval: "weekly"
-
-  - package-ecosystem: "swift"
-    directory: "/"
-    schedule:
-      interval: "weekly"
     target-branch: "dev"
+
+  # NOTE: no "swift" section. SPM packages live in the xcodeproj-embedded
```
Alexander5015: I think this is wrong because I haven't seen any errors with Dependabot with Swift, where did you see issues with it?

boringNotch/components/Notch/NotchHomeView.swift
```diff
@@ -202,7 +201,10 @@ struct MusicControlsView: View {
     }
 
     private var musicSlider: some View {
-        TimelineView(.animation(minimumInterval: musicManager.playbackRate > 0 ? 0.1 : nil)) { timeline in
+        // 0.5s ticks are imperceptible on a minutes-long track (~1px steps)
```
Alexander5015: Not true. This is not smooth (~1px steps) unless a song is 7–8 minutes long. The maximum interval this should be set to is at most 0.2s. 

While the 10 Hz cadence might be a bit aggressive as it is designed to have ~1px steps for songs 1.5 - 2 mins, we don't need to increase it if it isn't a massive drain since the view is only shown briefly.

boringNotch.xcodeproj/xcshareddata/xcschemes/boringNotch.xcscheme
Alexander5015, line 1: This shouldn't be committed

boringNotch/components/OSD/Views/SystemEventIndicatorModifier.swift
```diff
@@ -14,14 +14,15 @@ struct SystemEventIndicatorModifier: View {
     @Binding var value: CGFloat
     @Binding var icon: String
     @Binding var accent: Color?
-    let showSlider: Bool = false
+    /// Routed to hardware (volume/brightness) when the bar is dragged —
```
Alexander5015: This comment is probably unnecessary

Alexander5015 (dismissed):
Haven't had a chance to read everything because the diff is large, but here is a starting place

SECURITY.md
theboringhumane: Removed in 1f01e1e — table is gone, intro and reporting sections kept.

boringNotch/components/Notch/NotchHomeView.swift
theboringhumane: Agreed — renamed `WebcamPreview` → `WebcamView` so the type matches its file and the View-suffix convention (4060c04); call sites updated. `WebcamPreviewLayer` (the NSViewRepresentable) is unchanged.

.gitignore
theboringhumane: No good reason — the cleanup over-reached, and dropping `*.xcscheme` is exactly what let the scheme file get committed (your other comment). Reverted .gitignore wholesale to dev's version in 1f01e1e.

.swiftlint.yml
theboringhumane: Done — `.github/workflows/swiftlint.yml` added in 088f697 (checkout → brew install swiftlint → `swiftlint --config .swiftlint.yml`, on PRs to dev and stack/**). One note: it uses the standard `macos-26` label that cicd.yml's matrix already uses rather than the self-hosted `xcode-27` pin — one-word change if you'd rather have it there.

.swiftlint.yml
theboringhumane: Removed in 088f697.

.swiftlint.yml
theboringhumane: Removed in 088f697.

.swiftlint.yml
theboringhumane: Removed in 088f697.

.github/CODEOWNERS
theboringhumane: Fair pushback — to be clear on intent: this isn't meant as a territorial carve-out. The area list came from who historically touches what, and the default owner was moved to @theboringhumane deliberately since they've been driving the notification/XPC helper work heavily lately. It's routing, not a claim of sole expertise — your review is still the gate everywhere. Happy to reshape the mapping together (fewer areas, different split) rather than drop the mechanism entirely; what would you prefer?

.github/dependabot.yml
theboringhumane: Nowhere solid — the claim was wrong, and this repo's own history disproves it: the Sparkle 2.9.5 bump (#1434) came from Dependabot's swift ecosystem reading the xcodeproj-embedded manifest just fine. Restored the `swift` section and dropped the note in 1f01e1e.

boringNotch/components/Notch/NotchHomeView.swift
theboringhumane: Set to 0.2s in 4060c04 — you're right that ~1px only holds for very long tracks. The comment is corrected too: it now states 0.2s as the cap and credits the original 10 Hz design rather than claiming smoothness that isn't there.

boringNotch.xcodeproj/xcshareddata/xcschemes/boringNotch.xcscheme
theboringhumane: Untracked in 1f01e1e (`git rm --cached`, file stays on disk) — and since .gitignore is fully reverted, `*.xcscheme` is ignored again so it can't sneak back in.

boringNotch/components/OSD/Views/SystemEventIndicatorModifier.swift
theboringhumane: Removed in 4060c04.

PR #1511, opened 2026-09-02
feat : add menu bar

aglide100:


aglide100:
Ops, I accidentally submitted the PR here instead of to the forked repository.

Is it possible to delete it?

PR #1557, opened 2026-09-17
fix: allow album art click-through when idle

PersonalCao:
## Problem

When no music is playing, expanding the notch panel and clicking the album artwork does nothing — it should open the music app, but the click is silently swallowed.

## Cause

In `AlbumArtView.albumArtButton`, `albumArtDarkOverlay` is placed *after* the `Button` in the `ZStack`, so it renders on top of it. When playback is idle the overlay's opacity is 0.8 and, being an interactive SwiftUI view, it intercepts all hit tests, blocking taps from reaching the button underneath. While music is playing the overlay's opacity is 0, which masks the bug — clicks only work during playback.

## Fix

Add `.allowsHitTesting(false)` to `albumArtDarkOverlay` so taps always pass through to the album art button, regardless of playback state. The overlay is purely decorative, so this has no other behavioral effect.

One-line change in `boringNotch/components/Notch/NotchHomeView.swift`.

## Repro steps

1. Quit all media apps (nothing in Now Playing).
2. Expand the notch panel.
3. Click the album artwork → nothing happens (before the fix).

## Testing

- Built locally with Xcode 26.6 on macOS 26.5 (`xcodebuild -scheme boringNotch -configuration Release`) — build succeeds.
- Ran the app and verified: clicking the album artwork while idle now opens the music app; behavior during playback is unchanged.


Alexander5015:
Unable to reproduce this on latest `dev`. What I tried:
1. Clicking the Spotify icon (where album art would normally be) when Spotify is closed
2. Clicking the album art when music is playing
3. Clicking the album art when music is paused

All of the above worked for me, am I missing something?

PR #1558, opened 2026-09-17
feat: add QQ Music as a media source

PersonalCao:
## Motivation

QQ Music (com.tencent.QQMusicMac) is one of the most popular music apps among Chinese-speaking users. It has **no AppleScript dictionary**, so it can't be controlled the way Apple Music/Spotify are — but it does publish playback state to the macOS Now Playing channel. That makes it a natural fit for a dedicated media source: control goes through the system Now Playing channel (`NowPlayingController`), with QQ Music as the app to launch when the player is idle.

## Changes

- **`MediaControllerType`**: new `.qqMusic` case (settings picker and onboarding enumerate `allCases`, so the option appears automatically), plus a `usesNowPlaying` helper marking sources that rely on the system Now Playing channel.
- **`NowPlayingController`**: new `fallbackBundleIdentifier` init parameter (default `com.apple.Music`, so existing behavior is unchanged). It seeds the initial playback state and is used when a Now Playing update carries no bundle identifier — previously this fell back to an empty string, which left "open music app" without a target while idle.
- **`MusicManager`**:
  - `.qqMusic` maps to `NowPlayingController(fallbackBundleIdentifier: "com.tencent.QQMusicMac")`.
  - Now Playing-dependent sources (`.nowPlaying` and `.qqMusic`) share the same availability check, fallback, and runtime-recovery flow.
  - `openMusicApp()` resolves a sensible default app (QQ Music / Spotify / Apple Music per the selected source) when no bundle identifier is known yet. The `Defaults` read happens lazily at call time — reading it during initialization would deadlock against the `MusicManager.shared` singleton.
  - `togglePlay()` while the player is idle now opens the user's chosen app instead of sending a system play command (which makes macOS launch Apple Music regardless of the selected source).
- **Settings / onboarding**: the QQ Music option is disabled whenever Now Playing itself is unavailable (same treatment as the Now Playing option), and onboarding includes a description of the new source.

## Testing

- Built locally with Xcode 26.6 on macOS 26.5: `xcodebuild -project boringNotch.xcodeproj -scheme boringNotch -configuration Release build` succeeds.
- Ran the app and verified: the QQ Music option is selectable in Settings; clicking the album artwork or the play button while idle opens QQ Music; playback control (play/pause/next/previous, progress, artwork) works via the Now Playing channel while QQ Music is playing.

Happy to open an issue first for discussion if preferred.


Alexander5015:
Closing this because it just duplicates Now Playing. I will accept a PR for making Now Playing support filtering though, for a more general solution to this problem. The solution in my mind is something like whitelisting/blacklisting apps from a picker of some kind, but I haven't thought of a great UX here.

Issue #1571, opened 2026-09-21
[Bug] Poor scaling of music playback icons on custom notch height (under 24)


mrAnomalyy:
### What happened? What are the steps to reproduce the issue?

album art is scaled to unusable size (maybe replace with vector icon or remove completely at this size?)
music visualizer is getting cropped from top side

<img width="347" height="95" alt="Image" src="https://github.com/user-attachments/assets/c3c03c71-8593-49bb-a14c-fc9902ad1463" />

### What did you expect to happen?

replace music album cover with icon, correctly resize music visualizer

### Boring Notch Version

v2.7.3

### macOS Version

26.6.2

### Music Source? (If relevant)

Apple Music

### Relevant log output

```shell

```

### Checks

- [x] I haven't found any duplicates with my issue.

Alexander5015:
I think it looks ok the way it is on our latest beta release (the screenshot below is at 15pt). I set this to 15pt because I considered this to be the smallest usable size, so if 15pt doesn't work for some reason, I would much prefer to just bump up the minimum size. Visualizer could be adjusted though.

<img width="562" height="38" alt="Image" src="https://github.com/user-attachments/assets/995b7fbb-b931-4ffa-af55-260a706d4bc3" />

Alexander5015:
I'm gonna rewrite this issue a bit to target just the visualizer for now so I can track that

Alexander5015:
If you still feel like it is too small for album art tho in the new version, open a new issue for it

PR #1576, opened 2026-09-22
Hide music decorations in small notches

berkinefeavci:
Fixes #1571.

Small display cutouts can leave too little room for the music activity artwork and visualizer. Below 24pt, this change hides those decorations and returns no additional chin width, so the activity no longer creates empty horizontal space.

The existing behavior remains at 24pt and above. Added focused coverage for 15pt, 23pt, and 24pt.

Tests:
- standalone Swift compile check
- repository Python test suite: 18/18
- git show --check

Full Xcode tests were unavailable because this environment has Command Line Tools only.

AI assistance:
- Used AI assistance to investigate and prepare the change. I reviewed the diff and verification results before submitting.


Alexander5015:
I think it looks ok the way it is (the screenshot below is at 15pt). I set this to 15pt because I considered this to be the smallest usable size, so if 15pt doesn't work for some reason, I would much prefer to just bump up the minimum size. At numbers closer to 24pt (like 20pt) it looks even better. The padding on the left could be improved though

image

